PR #10, opened 2017-07-06
Media Queries Update and A-Z styling

kzadrozny:
Separated Media Queries from Print Styles. Added CSS for third party
A-Z styling.

### DIFF
--- a/base/static/base/css/_media.scss
+++ b/base/static/base/css/_media.scss
@@ -49,56 +49,5 @@
   }
 }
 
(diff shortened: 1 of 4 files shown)


kzadrozny:
@johnjung Can use the following paths for the Global Navigation bar on SFX's A-Z List:
base/static/base/css/base.scss
base/static/base/css/navigation.scss
base/static/base/css/third-party/azList.css

PR #44, opened 2017-09-20
Global Nav and Carousel Styling

kzadrozny:
Originally was going to add data models for the Reg Homepage, but decided to push this out separately.


kzadrozny:
Snuck the Law Library footer changes into this one too.

Issue #39, opened 2017-08-30
More Informative 404 and 500 Pages


kzadrozny:
Utilize Alex Gleason's [Wagtail Error Page](https://github.com/alexgleason/wagtailerrorpages) code to create more informative error pages on the library site.

Wagtail documentation on 404 pages here: http://docs.wagtail.io/en/v1.11.1/reference/project_template.html?highlight=404

kzadrozny:
Attempted to implement this last week, but we hit an issue with static files on my local machine. @bbusenius and I will work on this issue when we have extra time. Adding this comment so we can track how we fixed this issue for future use.

kzadrozny:
@bbusenius believes that upgrading Wagtail will resolve this issue, as they have made 404's more integrated.

PR #65, opened 2017-11-07
Custom 404

kzadrozny:
Allows conferences to be nested and is a patch to #39


kzadrozny:
Tested creating a nested Conference parent page on my local instance and everything worked well.

PR #109, opened 2018-11-16
Set the is_active field default to true and made migrations.

### LINKED ISSUE #98: Group page "is active" checkbox should be checked by default.
## Expected behavior
When creating a group page in loop the is_active checkbox should be checked by default. Since a user is creating the group, it follows that the group is active. This is the common use case.

## Actual behavior
When creating a group page in loop the is_active checkbox is not checked by default. If an author forgets to check it, the page will not be listed as an active group.

### Steps to reproduce
Create a group page for an "active group" but don't check the "is active" checkbox. The page will not be listed as an active group on the groups page: https://loop.lib.uchicago.edu/groups/

bbusenius:
Fixes #98


kzadrozny:
Is this failed check a false positive?

bbusenius:
I'm looking at it now. I'm pretty sure it's not related to the code change. Tests are just broken again. It seems to be a problem with redis. Either way, let's hold off on the merge until I figure out what's going on.

bbusenius:
Okay, it looks like this _was_ a false positive. I spent some time looking at this today. Tests always ran fine locally. When I ran them again in Travis, they passed. I'm not sure what triggered the initial failure but I don't think it's anything we need to worry about. This can be merged.

Issue #124, opened 2019-02-06
Loop News-Move 'Sticky' to Promote Tab


kzadrozny:
## Problem
UI issue with staff using "Stucky Until" improperly on the Loop News page

## Solution
Move "Sticky Until" date field from the main tab to the "Promote" tab so it is harder to accidentally fill in.

PR #132, opened 2019-02-12
Moved the sticky until field for loop news stories to the promote tab.

bbusenius:
Fixes issue #124.

Moved the sticky_until field to the promote tab.


kzadrozny:
Looks good to me. OK to go to production.

PR #152, opened 2019-05-23
Fix Unit Contact on Exhibit and Collection Pages

### LINKED ISSUE #151: Fix Unit Contact on Exhibit and Collection Pages
Both CollectionPage and ExhibitPage models have a `unit_page` BooleanField. When this field is selected in contact information should be pulled from the page's corresponding unit and displayed in the templates. Currently, nothing is displayed for ExhibitPages and the CollectionPage display is broken.

bbusenius:
Added a fix for unit_contact on exhibit and collection pages.

Fixes #151.

**Changes in this request**
Adds unit_contact display information to ExhibitPages and fixes the display on CollectionPages. Additional code cleanup.

### DIFF
--- a/lib_collections/models.py
+++ b/lib_collections/models.py
@@ -1,4 +1,5 @@
 from base.models import DefaultBodyFields, PublicBasePage
+from diablo_utils import lazy_dotchain
 from django.core.validators import RegexValidator
 from django.db import models
 from django.template.response import TemplateResponse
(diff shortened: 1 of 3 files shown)


lib_collections/models.py
```diff
@@ -329,72 +330,23 @@ def get_context(self, request):
...
-            try:
-                unit_link_page = self.unit.link_page.url
-            except:
-                pass
-
-            try:
-                unit_link_document = self.unit.link_document.file.url
-            except:
-                pass
+        unit_title = lazy_dotchain(lambda: self.unit.title, '')
```
bbusenius: @kzadrozny just for fun and FYI...

Python is an eager language. That means that arguments to functions are evaluated when they're passed to the function (not inside the function when they're used). This causes a problem because if any of these arguments such as `self.unit.link_document.file.url` don't exist, the code will blow up. That's why this code was previously in `try/except` blocks.

The `try/except` blocks cause the linter to complain because they're "bare". To fix them, they should be written as `except (AttributeError):`. I didn't want to do this because the code is repetitive. To simplify it, I wrote a function called `lazy_dotchain`. [This function is in my reusable library](https://github.com/bbusenius/Diablo-Python/blob/master/diablo_utils/diablo_utils.py) since I can use it on other projects. `lazy_dotchain` evaluates the existence of the arguments (e.g.`self.unit.link_document.file.url`) and returns the value if it finds it or a null value if it does not. In this case we set the null value to an empty string.

We get around python being an eager language by passing a `lambda`. Lambda functions are one of the few ways that python allows "laziness". In other words, since we pass the arguments to `lazy_dotchain` as `lambda` functions, they aren't evaluated until they're used inside `lazy_dotchain` instead of outside when they're passed. This allows me to set the `try/except` only once instead of multiple times like we did before.

kzadrozny:
This looks good to go and ready to merge.

PR #175, opened 2019-09-24
Updated StandardPages to pull news stories from Wagtail instead of WordPress

### LINKED ISSUE #174: Update kiosk pages to use internal news pages
## Description

`StandardPages` setup as _kiosk_ pages pull news stories via ajax from the news WordPress site. These pages should be updated to get news stories from the new news section of the site in Wagtail. Pages that use this functionality include:

- https://www.lib.uchicago.edu/
- https://www.lib.uchicago.edu/law/
- https://www.lib.uchicago.edu/crerar/
- https://www.lib.uchicago.edu/eck/
- https://www.lib.uchicago.edu/mansueto/
- https://www.lib.uchicago.edu/scrc/
- https://www.lib.uchicago.edu/ssa/

## Visual
image

bbusenius:
Fixes #174  

**Changes in this request**
Updated the code to source news stories internally from Wagtail and removed references to the WordPress news site.


kzadrozny:
@bbusenius What I can't easily tell from the diffs is if we are maintaining the duplication of news feeds for SSA and Eckhart? Eckhart gets Crerar's feed and SSA gets the Library feed.

base/models.py
```diff
@@ -41,13 +33,30 @@
...
     ('btn-primary', 'Primary'),
     ('btn-default', 'Secondary'),
     ('btn-reserve', 'Reservation'),
 )
 
+NEWS_CHOICES = (
+    ('', '--------'),
+    ('library_kiosk', 'Library'),
+    ('law_kiosk', 'Law'),
+    ('crerar_kiosk', 'Crerar'),
```
kzadrozny: EL requested this label be changed to "Sciences"

bbusenius: Right, so the way it will work is that on the Eckhart page they will choose the Crerar option from the dropdown. Previously it was a url field linking to Crear's feed. So, the real difference here is that it will be a dropdown in the Wagtail admin, instead of a UrlField. But, they'll still be able to use those options on their pages. So, Eckhart can have the Crerar feed and SSA can have the main Library feed.

kzadrozny: Great! I see this drop-down now in "News feed source" in the widgets tab. I dig this solution.

bbusenius: Ah yes, of course. I didn't change the link in the global navbar. Good catch! I'll change that and the Crerar label.

PR #189, opened 2019-10-02
188 improve news site feed performance

### LINKED ISSUE #188: Improve News Site Feed Performance
## Expected behavior
News site category pages should load quickly.

## Actual behavior
Despite caching, news site category pages are slow to load in some cases. This is true when a user is logged into the editor but can be true even when a user isn't logged in.

### Steps to reproduce
Go to a [news site category page](https://www.lib.uchicago.edu/about/news/category/business-economics/) while logged into the admin or continuously switch between category pages. You will see that the pages load slowly in some cases. 

## Other information:
Some of this has been improved and mitigated through the caching of page categories in the `LibNewsPage` method `get_categories` via `Redis` and by implementing [Wagtail Cache](https://docs.coderedcorp.com/wagtail-cache/stable/index.html). Nevertheless this has not been enough for a combination of reasons:
1. The django rest framework is _very_ slow.
2.  The call to `/api/v2/pages/` that our React app depends on is never cached when a user is logged in.
3. When a user isn't logged in, the call to `/api/v2/pages/` is cached, however, it gets moved out of the cache as other pages are cached and the cache reaches its limit.

bbusenius:
Fixes #188 

**Changes in this request**
- Sets up the news site feed so that it pulls from a minimized static JSON file rather than directly from the Django Rest Framework. The JSON file is updated when news stories are published or unpublished.
- Less egregious Redis caching of news site categories since this is no longer needed so much.


bufordrat:
Looks great to me, modulo the unit test issue we discussed.

PR #210, opened 2019-10-11
Delayed display of news category changes to published pages

### LINKED ISSUE #209: Delayed display of news category changes to published pages
## Expected behavior
When categories are changed or reordered on a public site news page and the page is published, the  changes should display immediately on the live version of the page.

## Actual behavior
Changes to categories display in a delayed fashion on the live version of the page even after the changes have been published.

### Steps to reproduce
Go to a news page and add another category to it or reorder the categories. The changes will not display until you empty the Redis cache.

## Cause and Solution
This is happening because the database queries for retrieving categories are cached in Redis. This problem can be solved by emptying the `news_cats` Redis cache in the `build_news_feed` function of `lib_news.models.py`. This is what we're already doing for the `wagtail_cache` template caching. This can be implemented by doing something like this:

```
from django.core.cache import cache
cache.delete('news_cats')
```

bbusenius:
Fixes #209 

**Changes in this request**
Adds the clearing of Redis cache for news site categories in the build_news_feed function. When a news site is published, the Redis cache for news categories is cleared, similar to what we do with the wagtail_cache.

### DIFF
--- a/lib_news/models.py
+++ b/lib_news/models.py
@@ -2,7 +2,7 @@ import json
 from urllib.request import URLError, urlopen
 
 import bleach
-from django.core.cache import caches
+from django.core.cache import cache, caches
 from django.db import models
 from django.template.defaultfilters import slugify
 from django.template.response import TemplateResponse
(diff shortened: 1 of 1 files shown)


bufordrat:
I can't reproduce this now (i.e. everything is working now), so I guess it's been sorted out.  But yesterday, for some reason, when I tried creating new news stories on my local machine in starfleet news, they would show up if they were the featured story, but not in the main list of news stories.  That is, if you toggled a new story off of being the feature story, it wouldn't then appear in the main list of news stories.  Anyway, since I can't seem to reproduce the issue right now, it probably isn't anything to worry about.  But if we see something like this happen again, we can remember this note.

PR #213, opened 2019-10-24
192 caching of views needed

bufordrat:
Fixes #.

**Changes in this request**

- 
- 
-

### DIFF
--- a/units/templates/units/unit_index_page.html
+++ b/units/templates/units/unit_index_page.html
@@ -30,7 +30,6 @@
           {% include "units/browse_by_pulldowns.html" %}
         {% endif %}
       </div><!-- / row -->
-
       {% if query %}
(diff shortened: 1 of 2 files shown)


bufordrat:
I added the @cache_page decorator to the collections, spaces, and staff directory views, but it doesn't seem to have an effect.  I think these views may already be getting cached.  See, for example, the following language, which suggests that the @cache_page decorator is for when you want to cherry pick which views get cached, rather than just caching everything you possibly can: 

"The wagtail-cache middleware will attempt to cache ALL responses that appear to be cacheable...
...To only cache specific views, remove the middleware and use the cache_page decorator on views or URLs."

From:
https://docs.coderedcorp.com/wagtail-cache/stable/getting_started/install.html

PR #214, opened 2019-10-24
168 add doi and pmid resolution service to wagtail search forms

### LINKED ISSUE #168: Add DOI & PMID resolution service to Wagtail search forms
See https://docs.google.com/document/d/16EMowZnwQftcKZuR-sYwqgn27qnJ84Rq07XNHERyGp4/edit?usp=sharing for details on service.

We will need to integrate the DOI/PMID resolution service into the search widget of Wagtail, applying it to search tabs on the home page and departmental kiosk pages, including Library Catalog, Articles, Journal, Database, and Website Search.

If the DOI/PMID resolution service detects that the user's search string is a valid identifier, it will supply an OpenURL to direct the user to. If the user did not enter a valid identifier, submit their search normally.

bufordrat:
Fixes #168 .

**Changes in this request**

This implements a feature whereby in all six of the search boxes on the main Wagtail site, if a user enters a well-formed DOI, instead of actually performing the relevant search (VuFind, Articles Plus, etc.) it obtains bibliographic information on the item corresponding to the DOI through CrossRef and redirects the user to the Find It! page for that record.  If the user enters an ordinary query that is not a DOI, the search box behaves normally--i.e. it searches the library catalog, or Articles Plus, or the library website, as the case may be.
- 
- 
-


public/utils.py
```diff
@@ -118,3 +120,140 @@ def has_feature(feature):
...
+    elif form_name == 'articles':
+        url = (
+            'http://proxy.uchicago.edu/login'
+            '?url=http://search.ebscohost.com/login.aspx'
+            '?direct=true&site=eds-live'
+            '&scope=site'
+            '&type=0'
+            '&mode=and'
+            '&cli0=FT1&clv0=Y'
+        )
```
bbusenius, line 248: I was unfamiliar with this way of doing a multi-line string in python. It seems to work fine in the interpreter. Cool!

public/utils.py
```diff
@@ -118,3 +120,140 @@ def has_feature(feature):
...
+    relevant search
+
+    Args: 
+        a string key indicating the type of form on the main page
+
+    Returns:
+        the URL to post to for the relevant search box
+    """
+    if form_name == 'catalog':
+        return 'https://catalog.lib.uchicago.edu/vufind/Search/Results'
```
bbusenius, line 238: Some of these URLs could _almost_ be set in the `base.py` config file. I say almost, because I don't think it really gets us much. I guess we could configure it to search dev instances of VuFind or something. That's not really something we need. This is really just a comment of the thinking out loud type. I'm fine with leaving them here as is. If we ever need to make this configurable, we can always do that in the future.

public/tests.py
```diff
@@ -1,3 +1,86 @@
...
+        ensure no exception gets thrown when trying to connect to
+        idresolve at the wrong URL
+        """
+        try:
+            doi_lookup_base_url(example_doi1, bad_url)
+            doi_lookup_base_url(example_doi2, bad_url)
+        except requests.ConnectionError as exception:
+            self.fail("public.utils doi_lookup raised %s"
+                      % str(exception)
+                      )
```
bbusenius, line 51: I like the way you handled this. This is a very good test and a nice way of handling the fail case in your corresponding code.

PR #217, opened 2019-11-18
215 page owners report for departed staff

### LINKED ISSUE #215: Page Owners report for departed staff
Rachel has requested a Page Owners Report for Yasmin Omer and Catherine Niehaus.

Yasmin Omer is no longer library staff, and she does not appear to be on the select list. She does still own some pages, though. Is it possible to allow inactive staff to be selected from the menu for reporting?

Also, I can select Catherine Niehaus from the reports interface, but I keep getting 504 Gateway Timeout messages when requesting this report.

bbusenius:
Fixes #215 

**Changes in this request**
-  Adds departed, inactive staff to the page owner's report.
-  Switches the page owner's report to `StreamingHttpResponse` instead of `HttpResponse` to help with the Gateway timeout problem.

### DIFF
--- a/base/forms.py
+++ b/base/forms.py
@@ -14,7 +14,7 @@ class PageOwnersForm(forms.Form):
 
     cnetid_choices = [
         (s.cnetid, s.title)
-        for s in StaffPage.objects.live().order_by('last_name', 'first_name')
+        for s in StaffPage.objects.all().order_by('last_name', 'first_name')
     ]
     label = 'Limit by staff member'
     cnetid = forms.ChoiceField(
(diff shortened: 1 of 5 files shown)


kzadrozny:
Code looks good, but I want to clarifying the workflow: This will still be a report that is run manually via Wagtail when we see that there is a departed staff via the directory diff cron job?

bbusenius:
Yep, that's correct. The only thing that really changes is that staff who have departed can now be selected. It's still a manual process.

Regarding the workflow, this hasn't been part of anyone's workflow as far as I know. It's just a reporting mechanism that can be used if someone wants it. It was originally requested for loop but we built it such that it would report on all sites.

bufordrat:
Very nice!

PR #234, opened 2020-02-03
198 rss feeds for the news site

### LINKED ISSUE #198: RSS feeds for the news site
## New feature request
RSS feeds are needed for the news section of the public site. A separate RSS feed should be created for every category on the site:

image

This should most likely be done as a Django `view` where category names are passed as part of a `GET` request.

## Not needed?
- There has been no preference expressed regarding the format of the urls
- There has been no request to have the feeds accessible as part of the templates

bufordrat:
Fixes #198.

**Changes in this request**
Separate RSS feeds for each news category.
When you navigate to the Library News main page, a link to an RSS feed for all library news appears in the bottom of the right sidebar.
When you navigate to any category of news, a link to an RSS feed for that category appears in the bottom of the right sidebar.
Each RSS feed displays the 20 most recent results.
- 
- 
-


base/templates/base/public_base.html
```diff
@@ -429,20 +429,46 @@ <h3>{{self.friendly_name}}Social Media</h3>
                                 {% endif %}
                             </ul>
                         </div>
-                    {% endif %}
-
-                    {% if libra %}
+			{% endif %}
+			
+			{% if libra %}
```
bbusenius: There are some indentation and whitespace problems in this template. This isn't the only place where it happens and I won't call them all out. It's not a big deal but we should fix it. Just block serves as a good example though. Line 432 has a closing if statement that doesn't line up with the opening if statement. 433 has empty whitespace on a newline and the html after line 434 is indented too much.

I don't believe we've always been consistent in all of our template files. Sometimes they use 4 space indentation and sometimes 2. I think 4 is more common on our site, and it's what's being used in this file, so let's go with that.

base/templates/base/public_base.html
```diff
@@ -429,20 +429,46 @@ <h3>{{self.friendly_name}}Social Media</h3>
...
+			<h3>
+			    Subscribe to {{ self.category }} <br>
+			</h3>
+			<div class="rightside-mod col-xs-12 col-sm-6 col-md-12">
+			    <a href="/rss/{{ self.slug }}/">
+				<i class="fa fa-rss-square" aria-hidden="true"></i>
+				RSS feed
+			    </a>
+			</div>
+			{% elif self.news_feed_api %}
```
bbusenius: The recommended Django best practice is to indent html in dynamic code blocks, even if it will ultimately render "over-indented" html in the browser. So, that means we would normally want to indent the html within the if clause. I can see we haven't done that consistently throughout this template (my bad) but we probably should.

base/templates/base/public_base.html
```diff
@@ -429,20 +429,46 @@ <h3>{{self.friendly_name}}Social Media</h3>
...
                         </div>
-                    {% endif %}
+			{% endif %}
 
-                    {% if contacts %}
+			{% if contacts %}
                         {% list_contacts_in_sidebar contacts 'Media Contact' %}
-                    {% endif %}
+			{% endif %}
+
```
bbusenius: Regarding the if `self.category` clause below, this only shows on LibNewsIndexPage, correct? There are cases where these code bits need to be in the public_base template. This might be one of those times, I'm not certain, however, I just want to be sure that we tried adding it to the lib_news_index_page template first. It's quite possible you tried this and it didn't work out. No worries if that's the case.

lib_news/models.py
```diff
@@ -1,6 +1,9 @@
 import json
 from urllib.request import URLError, urlopen
 
+from django.contrib.syndication.views import Feed
+from django.http import HttpResponse
```
bbusenius: I'm seeing linting errors in this file. These are mostly unused imports and whitespace errors.

lib_news/views.py
```diff
@@ -1,5 +1,11 @@
+from requests import get
```
bbusenius: Same as `models.py`, there are unused imports and whitespace errors.

lib_news/views.py
```diff
@@ -11,3 +17,61 @@ def ltdrfr(request):
...
+        """Generates the list of items in the feed."""
+        def has_category(cat):
+            def partial_application(page):
+                return cat in page.get_categories()
+            return partial_application
+
+        stories = (LibNewsPage
+                   .objects
+                   .order_by('-published_at')
+        )
```
bbusenius: E124: closing bracket does not match visual indentation (linting error)

lib_news/templates/lib_news/lib_news_index_page.html
```diff
@@ -71,13 +73,16 @@ <h1>Search results for: {{ self.search_query }}</h1>
...
-{% block right_sidebar %}
-    {% if display_current_web_exhibits %}
-        {% show_exhibits current_exhibits 'Current Exhibits' %}
-    {% endif %}
-{% endblock %}
+	{% if display_current_web_exhibits %}
+	{% show_exhibits current_exhibits 'Current Exhibits' %}
+	{% endif %}
+
+	{% endblock %}
```
bbusenius: Indentation

lib_news/templates/lib_news/lib_news_page.html
```diff
@@ -146,7 +146,7 @@ <h2>Visit the Exhibit</h2>
 
     {# Recent news stories #}
     {% if recent_stories %}
-        <section class="related-stories">
+    <section class="related-stories">
```
bbusenius: Indentation

base/templates/base/public_base.html
bufordrat: Right; this is only for LibNewsIndexPage.  I tried putting the html for the RSS link on the lib_news_index_page right-hand sidebar the first time around.  However, my understanding is that the public_base right-hand sidebar html gets inserted *after* the lib_news_index_page right-hand sidebar html.  David wanted the RSS feed link to appear at the bottom of the right-hand sidebar.  My sense was that those two factors were kinda forcing me to put the html for the RSS feed in the public_base template, as ugly as that is.  But maybe that's wrong.  What do you think?  It would be great to move this out of the public_base template if there's a nice way to do that.

base/templates/base/public_base.html
bufordrat: Ok, these tab/whitespace issues should be fixed in the latest commit.  Meanwhile, I'll work on setting my Emacs up to automatically indent Django templates according to our indentation spec.

base/templates/base/public_base.html
bbusenius: Yes, I'm sure that's right. I've had to do it with other modules as well. I just wanted to do due diligence and make sure I asked. It's fine as it is.

PR #279, opened 2020-03-25
Limited minute types to wagtail page and changed table styling

### LINKED ISSUE #255: Meeting Minutes Allow Only Wagtail Pages
Related to #50 : A short-term solution has been suggested until we have time to remodel the meting minutes - to remove the “external link” and “link document” options. This would leave only “Link Page” as an option for meeting minutes.

kzadrozny:
Closes #255 

**Changes in this request**
- Limited meeting minutes to only allow Wagtail pages
- Changed the minutes display table to a dictionary list

**Questions**
- Unsure if I should remove the else statements in the [meeting minutes model](https://github.com/uchicago-library/library_website/blob/master/group/models.py#L283) for determining between link and document files.
- Would love to switch the item link to be the date rather than the summary in the ["inner table"](https://github.com/uchicago-library/library_website/blob/master/base/templates/base/minutes_and_reports_inner_table.html#L6) but am not able to figure that one out.
- Also unsure why there is the `for item in items` line in this inner table since there is only one summary allowed per meeting minute item.

### DIFF
--- a/base/templates/base/minutes_and_reports_inner_table.html
+++ b/base/templates/base/minutes_and_reports_inner_table.html
@@ -1,10 +1,10 @@
-              {% for date, items in table.items %}
-                <tr>
-                  <td>{{date}}</td>
-                  <td>
+              <dl>
+                {% for date, items in table.items %}
+                  <dt>{{date}}</dt>
+                  <dd>
                   {% for item in items %}
-                    <a href="{{item.url}}">{{item.summary}}</a><br/>
+                    <a href="{{item.url}}">{{item.summary}}</a>
                   {% endfor %}
-                  </td>
-                </tr>
-              {% endfor %}
+                  </dd>
+                {% endfor %}
+              </dl>
--- a/base/templates/base/minutes_and_reports_inner_table.html
+++ b/base/templates/base/minutes_and_reports_inner_table.html
@@ -1,10 +1,10 @@
               <dl>
                 {% for date, items in table.items %}
                   <dt>{{date}}</dt>
-                  <dd>
                   {% for item in items %}
-                    <a href="{{item.url}}">{{item.summary}}</a>
+                    <dd>
+                      <a href="{{item.url}}">{{item.summary}}</a>
+                    </dd>
                   {% endfor %}
-                  </dd>
                 {% endfor %}
               </dl>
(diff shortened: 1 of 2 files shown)


group/models.py
```diff
@@ -142,10 +142,18 @@ class MeetingMinutes(AbstractReport):
     """
     Meeting minutes content type.
     """
-    panels = AbstractReport.panels
+    date = models.DateField(blank=False)
+    summary = models.TextField(null=False, blank=False)
```
bbusenius: I think these two lines can be removed. Since this class extends `AbstractReport` it should already have these fields and you should still be able to manually set the panels without them, the same as you're doing. The joys of inheritance!

group/models.py
```diff
@@ -142,10 +142,18 @@ class MeetingMinutes(AbstractReport):
...
+
+    panels = [
+        FieldPanel('date'),
+        FieldPanel('summary'),
+        PageChooserPanel('link_page'),
+    ]
 
     class Meta:
         abstract = True
+        ordering = ['-date']
```
bbusenius: I'll trust you on this one but you might want to check this on `nest` if you're not certain about it. It will effect ordering, I just don't remember exactly how. Assuming you've looked at this closely, disregard my paranoia.

base/templates/base/minutes_and_reports_inner_table.html
```diff
@@ -1,10 +1,10 @@
...
+                  <dt>{{date}}</dt>
+                  <dd>
                   {% for item in items %}
-                    <a href="{{item.url}}">{{item.summary}}</a><br/>
+                    <a href="{{item.url}}">{{item.summary}}</a>
                   {% endfor %}
-                  </td>
-                </tr>
-              {% endfor %}
+                  </dd>
```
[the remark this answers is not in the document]
kzadrozny: This construction is what I am unsure about:
why there is the `for item in items` line in this inner table since there is only one summary allowed per meeting minute item?

group/models.py
kzadrozny: Removed this, but added code to make linking to a page required. I didn't realize that the summary linked to nothing, even if no page was chosen!

group/models.py
kzadrozny: This was part of `AbstractReport`, but since it is inherited, I removed it.

base/templates/base/minutes_and_reports_inner_table.html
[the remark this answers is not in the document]
kzadrozny: It looks like you can have multiple `dd` without borking the list, so I've changed it to how you recommended.

Issue #280, opened 2020-03-26
Reusable content boxes in Wagtail


kzadrozny:
A rich text box in Wagtail that can be reused in several pages within Wagtail. If the "parent" box is updated, the updates cascade to all the pages the box is used in. We would like to replicate the same behavior that Springshare uses with their "shared boxes."

**Sketched Out Set-up**
- Use [snippets](http://docs.wagtail.io/en/v2.8/topics/snippets.html) to create the boxes
- A new model will be made: `ReusableContent`
- Add panel to Standard Page for "reusable content"
- Add panel to add "reusable content" so one can add an "alert widget" above Quicklinks
- It looks like we can use `wagtail.snippets.blocks.SnippetChooserBlock` so that the "reusable box" can be placed anywhere on the page ([related documentation](http://docs.wagtail.io/en/v2.8/topics/streamfield.html#snippetchooserblock))
- Snippet chooser should be repeatable so authors can populate a page with multiple reusable boxes (like in use for an FAQ)
- Add boolean for "add background color" to make the box stand-out more.

**Considerations**
Right now, only Admin folks can see and edit snippets. Are we able to change this?

**Time Estimate**
15-30 hours

kzadrozny:
Closed with branch #284

kzadrozny:
Reopening issue to add option for background color.
- Boolean for "make sidebar alert"
- Will add class `sidebar-alert` that will apply unique styling when reusable box is used as sidebar widget only

PR #284, opened 2020-03-28
280 Reusable content boxes in Wagtail

bbusenius:
Fixes issue #280 
 
**Changes in this request**
- Adds a new `ReusableContent` snippet that can be accessed in the admin. 
- Adds a new widget in the right sidebar that allows for adding reusable content. This field is repeatable so that multiple pieces of reusable content can be added.
- A new StreamFieldBlock for adding reusable content to the bodies of pages
- An [example page](http://wwwdev:8000/about/thelibrary/) with repeatable content in the right sidebar and the page body


bbusenius:
@kzadrozny I mocked up some example reusable content snippets and an example page (about page) in the test database. The example page has reusable content in the right sidebar and the top of the page body:
- http://wwwdev:8000/about/thelibrary/
If you should need the relevant html files are:
- `base/templates/base/blocks/reusable_content.html`
-  ` base/templates/base/public_base.html`

kzadrozny:
Will merge and push to production, once I get the OK from David.

PR #291, opened 2020-04-01
Added new alert category

kzadrozny:
Needs work. Stuck on a utils issue.

@bbusenius I am stuck on line 26 in alert.utils . See arrowed line with `alert`
```
else:
>      alert = high_alert[0] if high_alert else info_alert[0]
        msg = bleach.clean(
            alert.banner_message, tags=['p', 'b', 'a', 'strong'], strip=True
        )
        return (
            msg, alert.alert_level, alert.more_info,
            alert.relative_url(current_site)
        )
```

For this bleach section, I am unsure how to add a third option.

Note:
I still need to do styling for this new banner but I am currently unable to see it to do that! Once we fix this, I'll take it back to work on the styling.


bbusenius:
Ah, that should be straightforward. This line:

`alert = high_alert[0] if high_alert else info_alert[0]`

is setup as a _ternary operator_. You just need to refactor it as an `if`, `elif`, `else` statement where you set `alert` in each clause. I'll help you straighten this out tomorrow. This is a good exercise.

PR #303, opened 2020-04-13
298 update ask page template

### LINKED ISSUE #298: Update Ask Page Template
Changes are requested due to the streamlining of the Chat functions and to accommodate Springshare widgets.

Screen Shot 2020-04-08 at 9 55 42 AM

### Intro
- Add an "Intro" section
- Intro will be positioned before Chat
- Has field panels with limited streamfields:
   - H2-H4
   - Paragraph
   - Reusable Content
   - HTML

### Body
- Body remains, but is moved below Chat in template

### Consult a Specific Library
- Entire table is removed from template
Variations: Extra content can be moved to...
   - "Reference Resources" widget, which is a rich text box. Separate-looking modules can be created by using the `h3` and additional `ul`
   - Rich Text widget that is already being used on different template (reuse here)
Will look into these two options.

kzadrozny:
Closes #298 

**Changes in this request**
- Rearranged admin panels to follow published page order
- Moved email and contact form fields into one multipanel section since only one of those can be displayed at once
- Added Intro section
   - Opted not to add headers as separate streamfields since rich text covers headers
- Moved body section to fall below chat widget in template
- Put related links section in a new sidebar div in the template
   - Opted not to create a "new" widget in the admin; instead added extra help text about using h3s to create the look of a new sidebar element

**Note**
KZ needs to publish huge content changes when this goes live.


ask_a_librarian/models.py
```diff
@@ -16,9 +18,22 @@ class AskPage(PublicBasePage, ContactFields):
...
+        null=True,
+        blank=True,
+    )
     ask_widget_name = models.CharField(max_length=100, blank=True)
-    reference_resources = RichTextField(blank=True)
     body = StreamField(DefaultBodyFields())
+    reference_resources = RichTextField(
+        blank=True,
+        help_text=
+        'Links to guide links and other Ask pages. Make new sections with Header 3'
```
bbusenius, line 35: Nice addition!

ask_a_librarian/models.py
```diff
@@ -1,13 +1,15 @@
+from wagtail.core.blocks import RichTextBlock
 from django.core.validators import RegexValidator
 from django.db import models
 from wagtail.admin.edit_handlers import (
-    FieldPanel, MultiFieldPanel, PageChooserPanel, StreamFieldPanel
+    FieldPanel, FieldRowPanel, MultiFieldPanel, PageChooserPanel,
```
bbusenius, line 5: `FieldRowPanel` is not being used. We should probably just get rid of it.

ask_a_librarian/models.py
bbusenius, line 35: This isn't your fault, it's something annoying the autofixer is doing right now. We can get rid of the linting error here by doing something like:

```
    reference_resources = RichTextField(
        blank=True,
        help_text='Links to guide links and other \
        Ask pages. Make new sections with Header 3'
    ) 
```
After you format the text with the ecape `\` the autofixer will stop dropping it down to the new line.

PR #314, opened 2020-04-30
SMS Number with no migrations

### LINKED ISSUE #311: Switch Reference phone number to SMS number
While the building is closed, we don't want to be promoting the voice phone number for the Reference Desk. At the same time, reference staff wish to pilot an SMS-based chat service. We should replace occurrences of the voice number with the SMS number.

Also, clearly indicate that the new number is for texting with a label of "Text us" or "SMS Text Reference" and an icon of a cell phone (in contexts where an icon needs to appear).

Pages where this number occurs:
https://www.lib.uchicago.edu/research/help/ask-librarian/
https://www.lib.uchicago.edu/about/directory/ (in sidebar)
"Research and Teaching Support" unit on https://www.lib.uchicago.edu/about/directory/?view=department

LibGuides owned by the "Ask a Librarian" account.
Ie: http://guides.lib.uchicago.edu/articles

kzadrozny:
Closes #311 

## Note
- No migrations have been made yet so as not to clash with the Wagtail upgrade

## Problems (Need Help)
- Issue in trying to update the "Quick Numbers" section in the sidebar of [the Directory view](https://www.lib.uchicago.edu/about/directory/).
   - This is referenced as `quick_nums` in [unit_index_page.html](https://github.com/uchicago-library/library_website/blob/master/units/templates/units/unit_index_page.html#L118), which then refers to the `location` field in a [Unit Page in Loop](https://www.lib.uchicago.edu/admin/pages/2455/edit/#tab-human-resources-info).
   - The `location` field is a selector for a [Spaces page](https://www.lib.uchicago.edu/admin/pages/1797/edit/#tab-content) in Public. This page has a field for number and number label, but neither of these display in the Quick Numbers sidebar of the Directory.
   - I have no idea where the other numbers and labels are being pulled from as a search of the site yields nothing nor are these numbers on those respective department's pages.
- I believe / hope the "Quick Numbers" is data that can be changed within Wagtail once we locate where it is being pulled from, which will mean that no code changes will be needed to change this bit.
- Once we locate where the quick numbers are pulled from, I highly suggest Documentation is written for this widget.

## Changes Made
- Changed label for Secondary number to SMS Number.
- Changed icon for SMS Number
- Updated icons with more modern ones for the Ask page
- Made the icons uniform in size on both the Ask page and the Sidebar of the Directory Browse


bbusenius:
The quick numbers are part of our config. You can find them in the `QUICK_NUMS` setting of `library_website/settings/base.py`.

A few things to note about this:
1. Ignore the linting errors here. We wrote it in violation of linting errors on purpose to improve readability.
2. Know that it is at least an option to copy the `QUICK_NUMS` dictionary into `local.py` and alter it there instead of `base.py`. You would want to do this if you don't expect the quick number changes to be permanent, as that will be the easiest to revert.
3. We have a few unit tests that test the integrity of the quick numbers config/code. We'll want to be sure these pass with the new configuration. I expect they will but if you want to be sure, just run `./manage.py test` after you make the changes and go get some coffee.

ask_a_librarian/models.py
```diff
@@ -38,7 +38,10 @@ class AskPage(PublicBasePage, ContactFields):
     )
     phone_regex = RegexValidator(regex=PHONE_FORMAT, message=PHONE_ERROR_MSG)
     secondary_phone_number = models.CharField(
-        validators=[phone_regex], max_length=12, blank=True
+        validators=[phone_regex],
+        max_length=12,
+        blank=True,
+        verbose_name="SMS Number"
```
bbusenius, line 44: Using `verbose_name` was a nice, low-impact way to do this 👍

PR #316, opened 2020-05-05
315 upgrade to wagtail 2.9 and django 3.0.6

### LINKED ISSUE #315: Upgrade to Wagtail 2.9 and Django 3.0.6
It's better and faster to upgrade now, having just done the upgrade to 2.8.1. Having looked at it closely, I've assessed that this is a very simple update.

bbusenius:
Fixes #315 

**Changes in this request**
- Fixes Wagtail code deprecations
- Removes unused code
- Upgrades to `Wagtail 2.9` and `Django 3.0.6`


base/views.py
```diff
@@ -20,25 +20,6 @@
 from units.utils import get_default_unit
 
 
-def breadcrumbs(request):
```
bufordrat, line 23: Nice!

bbusenius: Yeah, it's weird that that code wasn't being used except by that view which also wasn't being used. I didn't trust it at first but that's what I determined. By all means if you understand differently, let me know!

PR #342, opened 2020-05-21
Schedule Appointment Button in Featured Expert Widget

### LINKED ISSUE #338: Schedule Appointment button to expert widget
Add a "Schedule Appointment" button to the widget, as in the second option here: https://docs.google.com/presentation/d/1bYUUTF-d8OLCBtc41fZlO8VLUr3kizKL1AtHnSA1RoA/edit?usp=sharing

## Related Code
```
<div class="col-xs-4 home-profile">
   <img src="{{profile_pic.url}}" alt="{{profile_pic.title}}" class="img-responsive" />
</div>
```

bufordrat:
Fixes #338.

**Changes in this request**
- add a 'Schedule an Appointment' to Featured Expert Widget 
- button is the same as the one that appears in the staff directory listing
- and also the same as the one that appears on each staff member's staff page


library_website/settings/base.py
```diff
@@ -33,11 +33,9 @@
     'django.contrib.messages',
```
bbusenius, line 33: Am I correct that this file (`base.py`) and `staff/models.py` only have linting changes? If that's the case could we revert these? It will help cut down on some of the unrelated noise and troubleshoot the problem with unit tests. If I'm wrong about that, please disregard.

public/templates/public/blocks/featured_library_expert.html
```diff
@@ -23,10 +23,8 @@ <h6>Not sure where to start?</h6>
             <li><a href="{{self.url}}">Ask a Librarian</a></li>
         </ul>
     </div>
-    {% if profile_pic %}
-        <div class="col-xs-4 home-profile">
-            <img src="{{profile_pic.url}}" alt="{{profile_pic.title}}" class="img-responsive" />
-        </div>
-    {% endif %}
+
```
bbusenius, line 26: If this spacing is showing correctly in git (sometimes it doesn't) we probably don't need this much.

staff/utils.py
```diff
@@ -806,3 +806,10 @@ def lookup_staff_ids():
         return {person['email']: person['id'] for person in json}
     except(JSONDecodeError, requests.exceptions.MissingSchema):
         return ''
+
+
+def libcal_id_by_email(emailaddr):
+    try:
+        return lookup_staff_ids()[emailaddr]
```
bbusenius, line 813: Looks like this line is causing the unit tests to fail. It would be easy to solve this by handling for `TypeError` as well but you'll want to make sure this doesn't cause any other problems.

bufordrat:
> It looks like `test_file_for_slack` is a mistake. We should get rid of that. We'll also need to fix the problem with unit tests. I left a comment about what I think is causing them to fail.

Sorry--is that a file somewhere in the project?  When I `grep` for it, I get the following:

```
> grep --exclude=\*.{json, js} -ri 'test_file_for_slack'
Binary file .git/index matches
```

library_website/settings/base.py
bufordrat: Yes; that's right.  They used to have substantive changes, but then I took them out, and I guess the linting changes are what's left.  Reverting them now.

bufordrat: Ok; those should be back to normal.

bbusenius:
`test_file_for_slack` is showing in the diff: https://github.com/uchicago-library/library_website/pull/342/files

public/templates/public/blocks/featured_library_expert.html
bufordrat: Ah, weird.  I'm going to go ahead and revert this to the `master` version for now since we're planning to open up a new issue anyway investigating whether it's getting used.

bufordrat:
> `test_file_for_slack` is showing in the diff: https://github.com/uchicago-library/library_website/pull/342/files

Oh, duh.  I see; it's the name of the file, so no point in trying to find it with `grep`.  Ok, removing it now!

staff/utils.py
bufordrat: Thanks for pointing this out.  I think I have more of a feel for how to interpret the tea leaves of a failed unit test now.

bufordrat: I'll handle a `TypeError`, then go in and see whether this noticeably messes anything up on the Wagtail site.

bufordrat: (Though I'd like to know how the unit test ended up passing `libcal_id_by_email` an int or whatever it was.)

bufordrat: I haven't been able to find any problems with also catching the `TypeError` so far, so I guess I'll go-ahead and push the change up to the branch.

kzadrozny:
The `if` statement is working as expected for scheduler button. Reminder to clear Wagtail cache after pushing to production.

PR #384, opened 2020-09-15
Added drop down style options

### LINKED ISSUE #383: Add Information styling for Reusable Box Alert
Requested from Communications: the alert box color could be the calming light blue of the current banner alert, rather than the pink/red of the current in page alert. This could either be done by globally changing the current “in page alert“ style or adding a second, lower tier style as another option.

kzadrozny:
Closes #383 

**Changes in this request**
- Added drop down style options for alert box
- Fixed styling issues with widget in sidebar

## Side effect
All snippets that had check box of "make alert" will lose this styling. Kathy will need to make a list of which snippets have this styling and apply the new field to them after this is pushed to production.


bbusenius (approved):
These are very nice and concise code changes. It looks exactly how I would have expected. I like how you got all the migrations into one file and how these changes simplify the template code.

Because of the possible ramifications of deploying this onto a site with lots of data already in it, this is one where I recommend deploying and testing on `nest` first, before putting it on production. Also, I notice our CI is broken again, since, Aug 13th. I'm not sure what that's all about but I doubt that's anything you did. I'll add that to my list of things to investigate.

kzadrozny:
Code test on Nest with local tests run by Brad.

PR #391, opened 2020-09-17
 The loop-search view is accessible from the public domain

### LINKED ISSUE #386: The loop-search view is accessible from the public domain
## Expected behavior
The [loop-search view ](https://loop.lib.uchicago.edu/loop-search/?query=library+templates)should _never_ be accessible to anyone other than a Library staff member who has authenticated.

## Actual behavior
This view is currently accessible to unauthenticated users from the public site domain, e.g.: https://www.lib.uchicago.edu/loop-search/?query=library+templates. This should not be possible. We need to prevent this from happening. The loop-search view should respect the same privacy settings as loop pages.

bbusenius:
Fixes #386 

**Changes in this request**
- Moved the `loop_homepage` variable definition to the top of the view and added the permissions checking code we use on normal loop pages.
- Linted the file

Linting makes it look like many changes were made, however, all I really did is move the `loop_homepage` variable up and add this blurb:

```
    if not has_permission(request.user, get_required_groups(loop_homepage)):
        return redirect_users_without_permissions(
            loop_homepage, request, None, None
        )
```

### DIFF
--- a/search/views.py
+++ b/search/views.py
@@ -1,12 +1,21 @@
+from base.wagtail_hooks import (
+    get_required_groups, has_permission, redirect_users_without_permissions
+)
+from django.core.paginator import EmptyPage, PageNotAnInteger, Paginator
 from django.shortcuts import render
-from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
-from intranethome.models import IntranetHomePage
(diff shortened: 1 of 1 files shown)


kzadrozny (approved):
Looks good to me!

PR #393, opened 2020-09-18
News stories without excerpts blow up the news homepage

### LINKED ISSUE #387: News stories without excerpts blow up the news homepage
## Problem
When a news story does not have an excerpt field (filled out in the admin), the news homepage fails to load with a `TypeError`. This is because the method responsible for creating a "short description" is failing when the body content is passed as a fallback because an excerpt is not present. It fails because the method expects a `string` but receives a `StreamValue` instead. 

### Steps to reproduce
In dev:
1. Remove the excerpt for the feature story of the news site by deleting it in the admin and publishing the page. 
2. Go to the news homepage

### Related code
The `short_description` method on line 400 in `lib_news/models.py` needs to be refactored to handle `StreamValues` or the fallback needs to be removed.
```
    @property
    def short_description(self):
        if self.excerpt:
            retval = self.excerpt
        else:
            retval = self.body
        # return retval
        return bleach.clean(
            retval,
            tags=['a', 'b', 'i'],
            attributes={'a': ['href', 'rel', 'data', 'id', 'linktype']},
            strip=True,
            strip_comments=True,
        )
```

bbusenius:
Fixes #387

**Changes in this request**
- Changes to the body fallback for, "short descriptions". If an excerpt is no longer present for a news story, short descriptions are created by truncating html from the news story body. This fallback will most likely never be used for an extended period of time in reality, since, this bug only appears to affect feature stories (that always have excerpts). In all likelihood this just protects the site while feature story author's are working on their stories. 
- Feature stories without excerpts no longer crash the news site homepage.
- The `get_first_feature_story` now only returns `live` published stories.

There are no migrations with this change set because the updates only affect a function that alters data after it leaves the database. It doesn't affect anything that is re-stored in the database.

### DIFF
--- a/lib_news/models.py
+++ b/lib_news/models.py
@@ -6,7 +6,7 @@ from django.core.cache import cache, caches
 from django.db import models
 from django.template.defaultfilters import slugify
 from django.template.response import TemplateResponse
-from django.utils import timezone
+from django.utils import text, timezone
 from modelcluster.fields import ParentalKey
 from rest_framework import serializers
 from wagtail.admin.edit_handlers import (
@@ -401,7 +401,8 @@ class LibNewsPage(PublicBasePage):
         if self.excerpt:
             retval = self.excerpt
         else:
-            retval = self.body
+            html = str(self.body).replace('</p>', '</p> ')
+            retval = text.Truncator(html).words(100, html=True)
         # return retval
         return bleach.clean(
             retval,
--- a/lib_news/utils.py
+++ b/lib_news/utils.py
@@ -6,7 +6,7 @@ def get_first_feature_story():
         LibNewsPage object.
     """
     from .models import LibNewsPage
-    feature = LibNewsPage.objects.filter(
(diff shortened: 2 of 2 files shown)


kzadrozny (changes requested):
If a new story is marked as a feature and saved as a draft it is still published to the news home page.
Additionally, when a feature story is published and then unpublished the feature remains on the news homepage. This second error may be a caching issue.

bbusenius:
**Nice catch!** Actually, both of those issues are the same thing. It's just displaying the first feature story and disregarding the fact that it might be unpublished. It is definitely a separate issue but it's connected to the original problem. My changes couldn't have caused it, which means, this problem is _also_ affecting production. I patched it. That actually explains a lot, because it was probably a draft that R was working on when the short description/no excerpt problem came up (since R always provides excerpts for feature stories). I can't believe that problem existed on the live system all this time. I'm very glad you discovered it.

kzadrozny (approved):
Working as expected!

PR #416, opened 2020-10-23
Fix libguides ebook search

### LINKED ISSUE #415: Fix libguides ebook search
## Expected behavior
Searches for ebooks using [the ebooks search form on this libguide ](https://guides.lib.uchicago.edu/c.php?g=297099&p=2011186#s-lg-box-6061581)should render a search in the Library catalog.

image

## Actual behavior
Searches using this form fail because the old php form processing script no longer exists. It was phased out when we migrated to Wagtail.

### Steps to reproduce
Go to the form using the link above and attempt a search.

bbusenius:
Fixes #415 

**Changes in this request**
- Adds a new view that takes a query and forwards it on as an ebooks search in the catalog.


kzadrozny (approved):
Works on my local machine as expected. VuFind is a little slow to load, but the facets are active on the search when the results come through.

PR #418, opened 2020-10-26
Unit tests failing due to change in University directory xml

### LINKED ISSUE #417: Unit tests failing due to change in University directory xml
## Problem
Unit tests are failing when we run `./manage.py test` and in Travis CI. This appears to be due to a superficial change in the directory xml provided by the University. As far as I can tell, this change does not effect our code. It only causes the test to fail.

bbusenius:
Fixes #417 

**Changes in this request**
- Changed the XML DTD in tests to match the XML being returned by the University API. Tests should pass now.


kzadrozny (approved):
Editing, saving as draft, and publishing Loop and Public pages work without any errors. Was able to run unit sync cron commands without errors.

PR #438, opened 2021-01-07
Added alt text to duo image

### LINKED ISSUE #437: SiteImprove ADA Issues
Small issues to fix automated ADA errors from SiteImprove
- Make empty `alt` always show in `img`
- Change if statement to be text within `alt` within `img`
- Change footer `li` [role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role) from `header` to `heading`
- Review Loop teamplates to see which one doesn't define language
-  Make link color in alert darker (color contrast)

kzadrozny:
Closes #437

**Changes in this request**
- Added alt text to duo images
- Changed `header` to aria correct term `heading` in footer list

### DIFF
--- a/base/templates/base/blocks/duo_img.html
+++ b/base/templates/base/blocks/duo_img.html
@@ -4,7 +4,7 @@
 <div class="col-xs-12 duo-wrapper">
     <div class="col-xs-12 col-sm-6 collex-duo">
         <figure>
-            <a href="{{original_one.url}}" data-toggle="lightbox"><img class="img-responsive" src="{{original_one.url}}" alt="{% firstof value.alt_text value.image_one.alt_text %}"></a>
+            <a href="{{original_one.url}}" data-toggle="lightbox"><img class="img-responsive" src="{{original_one.url}}" alt="{{value.image_one.alt_text}}"></a>
             <figcaption>
                 <span class="img-title">{{value.image_one.image}}</span><br>
                 <span class="img-citation">{{value.image_one.citation}}</span>
@@ -14,7 +14,7 @@
     </div>
     <div class="col-xs-12 col-sm-6 collex-duo">
         <figure>
-            <a href="{{original_two.url}}" data-toggle="lightbox"><img class="img-responsive" src="{{original_two.url}}" alt="{% firstof value.alt_text value.image_two.alt_text %}"></a>
+            <a href="{{original_two.url}}" data-toggle="lightbox"><img class="img-responsive" src="{{original_two.url}}" alt="{{value.image_two.alt_text}}"></a>
             <figcaption>
                 <span class="img-title">{{value.image_two.image}}</span><br>
                 <span class="img-citation">{{value.image_two.citation}}</span>
--- a/base/templates/base/blocks/duo_img.html
+++ b/base/templates/base/blocks/duo_img.html
@@ -4,7 +4,7 @@
 <div class="col-xs-12 duo-wrapper">
     <div class="col-xs-12 col-sm-6 collex-duo">
         <figure>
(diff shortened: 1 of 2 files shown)


base/templates/base/blocks/duo_img.html
```diff
@@ -4,7 +4,7 @@
 <div class="col-xs-12 duo-wrapper">
     <div class="col-xs-12 col-sm-6 collex-duo">
         <figure>
-            <a href="{{original_one.url}}" data-toggle="lightbox"><img class="img-responsive" src="{{original_one.url}}"></a>
+            <a href="{{original_one.url}}" data-toggle="lightbox"><img class="img-responsive" src="{{original_one.url}}" alt="{% firstof value.alt_text value.image_one.alt_text %}"></a>
```
bbusenius, line 7: I think it might be impossible to populate this with `value.alt_text` or maybe I just forgot how to do it. If I'm right, we should just use `value.image_one.alt_text` and abandon the `firstof` tag. Does that make sense? Let me know if I just forgot how to do this.

base/templates/base/blocks/duo_img.html
```diff
@@ -14,7 +14,7 @@
     </div>
     <div class="col-xs-12 col-sm-6 collex-duo">
         <figure>
-            <a href="{{original_two.url}}" data-toggle="lightbox"><img class="img-responsive" src="{{original_two.url}}"></a>
+            <a href="{{original_two.url}}" data-toggle="lightbox"><img class="img-responsive" src="{{original_two.url}}" alt="{% firstof value.alt_text value.image_two.alt_text %}"></a>
```
bbusenius, line 17: Same as above.

bbusenius (changes requested):
I think we might need to make a minor change. See my comment in the code.

base/templates/base/blocks/duo_img.html
kzadrozny: Without the `firstof` the site blows up. I copied this structure from the solo image template, which is using `firstof`. Let me know if there is a different way I would need to structure this if `firstof` is problematic.

base/templates/base/blocks/duo_img.html
kzadrozny: See above: does not work without `firstof` in the current structure.

base/templates/base/blocks/duo_img.html
kzadrozny: Additionally: removing `value.alt_text` populated the alt text area with weird code instead of the actual text in the data field. It was bizarre.

bbusenius: Without `firstof` you would just do it like this `alt="{{value.image_one.alt_text}}"`, I think (notice `{{` instead of `{%`). I know when I've done the `firstof` technique in the past, I was throwing values at it that were populating from different fields in order of preference. I just can't for the life of me figure out how to populate `value.alt_text` and even if I could, I think it would be the same for both images. In order to see this concretely, I just dumped their values at the top of the template in `p` tags along with those for `value.image_one.alt_text` (and `image_two`) just so I could see them when I tried populating them in the admin. You might be getting an error because you did something like this `{% value.image_one.alt_text %}`. That's my guess.

kzadrozny: This has been updated to `alt="{{value.image_one.alt_text}}`

base/templates/base/blocks/duo_img.html
kzadrozny: This has been updated to `alt="{{value.image_two.alt_text}}`

bbusenius (approved):
Looks good!

PR #453, opened 2021-01-25
Added if statement to closing date

### LINKED ISSUE #452: Exhibits News Widgets closing date error
## Problem
If closing date left blank on exhibit parent page, the Exhibits News Widgets displays `None`

## Fix
Wrap closing date in `if` statement so is it blank if no closing date provided.

kzadrozny:
Closes #452

**Changes in this request**
- Included closing date into if statement for opening and closing date.

### DIFF
--- a/public/templates/public/blocks/related_exhibit.html
+++ b/public/templates/public/blocks/related_exhibit.html
@@ -14,7 +14,9 @@
       <br/>{{self.exhibit.exhibit_open_date}}
       {% if self.exhibit.exhibit_open_date and self.exhibit.exhibit_close_date %}
         —
-      {{self.exhibit.exhibit_close_date}}
+      {% endif %}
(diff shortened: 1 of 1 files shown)


bbusenius (approved):
This looks fine to me. Open dates will always display if they exist and close dates will only display if there is also an open date. I don't feel like this is problematic. One thing to note, however, is that if an open date is _not_ provided it might display `None` the same way the empty close date did. I feel like this would have come up by now if that scenario was likely. 

An alternative way to have done this would have been to wrap both `self.exhibit.exhibit_open_date` and `self.exhibit.exhibit_close_date` in their own if statements and leave the other if statement the way it was. That might have been overkill, I'm not sure.

kzadrozny:
Kept hyphen in original `if` statement and moved the closing date into its own `if` statement to allow for any future exhibits that only have a closing date.

PR #461, opened 2021-02-26
Fixed small ADA automation issues

### LINKED ISSUE #449: SiteImprove Automated Medium Fixes
## Empty heading tags
**Problem**
Authors are leaving empty heading tags by accident in the content of their pages. This leads to automated ADA compliance errors.
**Fix**
Add code to strip empty header tags from the DOM.

## Solo + Duo Image caption color contrast
Make solo and duo image streamfield captions be a solid stable color, not one that replies on the color scheme of web exhibits, to assure consistent color contrast and readability. (Example page: [Envisioning South Asia](https://www.lib.uchicago.edu/collex/exhibits/envisioning-south-asia-texts-scholarship-legacies/covering-modernity/))

## Carousel Issue when 1 item
**Problem**
When only one image is loaded into the carousel, `role="listbox"` still shows up and causes ADA compliance errors.

**Fix**
Have `listbox` only show up in multiple images are loaded into the carousel.

## Element IDs are not unique 
- multiple `#event-header` in Events homepage module (each event item has the id)
   - Also occurs anywhere else that Events module is (eg: News pages)

## Text is clipped when zoomed in
- "All fields" etc drop down on main search box is said to clip text when zoomed in.
- Cannot reproduce issue, but SiteImprove suggests we add `overflow: auto;` to the div

kzadrozny:
Closes #449 

**Need Help on**
- Changed `#event-header` to `.event-header` for events sidebar module js.
- Code didn't seem to need this to be an ID, so changed to class to there aren't multiple duplicate ids when multiple events
- Unsure how to test: Could not verify this didn't break event module because local instance is fake and Nest is currently frozen.

**Changes in this request**
- Added css to hide empty header tags
- Switch Sass function to darken citation text color (instead of lighten)
- Added if statement to `role-"list-item"` in carousel module
- Changed oveflow in search filter select item to `auto` instead of bootsrap default. Didn't seem to break anything in zoomed in view.

### DIFF
--- a/base/static/base/js/library_website.js
+++ b/base/static/base/js/library_website.js
@@ -17,10 +17,10 @@ function renderEvents() {
             $.each(innerJson, function(i, v){
                 if (v['start_date'] == v['end_date']) {
                     // single-day events.
-                    eventsHtml += '<p><a id="event-header" href="' + v['link'] + '">' + v['title'] + '</a><br/><span class="event-date">' + v['start_date'] + '</span> | ' + v['start_time'] + ' - ' + v['end_time'] + '</p>'
+                    eventsHtml += '<p><a class="event-header" href="' + v['link'] + '">' + v['title'] + '</a><br/><span class="event-date">' + v['start_date'] + '</span> | ' + v['start_time'] + ' - ' + v['end_time'] + '</p>'
                 } else {
                     // multi-day events. 
-                    eventsHtml += '<p><a id="event-header" href="' + v['link'] + '">' + v['title'] + '</a><br/><span class="event-date">' + v['start_date'] + '</span> | ' + v['start_time'] + ' -<br/>' + '<span class="event-date">' + v['end_date'] + '</span> | ' + v['end_time'] + '</p>'
+                    eventsHtml += '<p><a class="event-header" href="' + v['link'] + '">' + v['title'] + '</a><br/><span class="event-date">' + v['start_date'] + '</span> | ' + v['start_time'] + ' -<br/>' + '<span class="event-date">' + v['end_date'] + '</span> | ' + v['end_time'] + '</p>'
                 }
             });
             $('#events-target').replaceWith(eventsHtml);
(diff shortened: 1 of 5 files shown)


base/static/base/js/library_website.js
```diff
@@ -17,10 +17,10 @@ function renderEvents() {
...
                 if (v['start_date'] == v['end_date']) {
                     // single-day events.
-                    eventsHtml += '<p><a id="event-header" href="' + v['link'] + '">' + v['title'] + '</a><br/><span class="event-date">' + v['start_date'] + '</span> | ' + v['start_time'] + ' - ' + v['end_time'] + '</p>'
+                    eventsHtml += '<p><a class="event-header" href="' + v['link'] + '">' + v['title'] + '</a><br/><span class="event-date">' + v['start_date'] + '</span> | ' + v['start_time'] + ' - ' + v['end_time'] + '</p>'
                 } else {
                     // multi-day events. 
-                    eventsHtml += '<p><a id="event-header" href="' + v['link'] + '">' + v['title'] + '</a><br/><span class="event-date">' + v['start_date'] + '</span> | ' + v['start_time'] + ' -<br/>' + '<span class="event-date">' + v['end_date'] + '</span> | ' + v['end_time'] + '</p>'
+                    eventsHtml += '<p><a class="event-header" href="' + v['link'] + '">' + v['title'] + '</a><br/><span class="event-date">' + v['start_date'] + '</span> | ' + v['start_time'] + ' -<br/>' + '<span class="event-date">' + v['end_date'] + '</span> | ' + v['end_time'] + '</p>'
```
bbusenius, line 23: This is a low risk change. This is just the html rendered on the page after the API call. Since you changed the SASS to match, I'm inclined to think this will be OK.

bbusenius (approved):
This looks good. I was not able to test either, however, this is a pretty low risk change. The test db is using an outdated url for the events widget. This can be switched by copying the one from production (wicket) and plugging it in, however, this will still not work because wicket is locked down to campus. I should think you could test this over the VPN if you use it. If that doesn't work, nest or another campus dev box are the only other options. Normally I would do this over a proxy, however, I'm unable to make this work with the dev environment. I can go to wicket but I cannot make it work with the site. I'm sure there's a way to do it but it's eluding me at the moment.

PR #475, opened 2021-03-03
Aria label for lightbox images

### LINKED ISSUE #460: Exhibit Images: SiteImprove ADA Issues
Multiple automated ADA issues with Solo / Duo image stream fields.

## No alt text
- Getting error of no alt text even when `figcaption` is filled in.
- Multiple way to move forward on this. One is to pull a unique ID for each solo / duo div and add a `aria-described-by` tag when caption is filled.
- May also need to simplify `figcaption` section and remove citation area into its own div. (More research needed.)

## Link without a text alternative
- All exhibits for solo / duo images are erroring out for no alt text. Switch from having caption required to having alt text required

kzadrozny:
Closes #460

**Changes in this request**
- Added aria label within link tag for lightbox images. Fixes SiteImprove error re: link without text.
- Changed alt text help text field to be more explanatory
- Made alt text in ImageBlock required
- Bumped contrast on "view all" links and added focus styling
- Added focus styling to "Schedule Appointment" button


bbusenius (changes requested):
This looks good. I don't see any problems in the code. We just need to add fixtures.

bbusenius (approved):

PR #482, opened 2021-03-25
Removed name requirements for reports pages

### LINKED ISSUE #481: Reports Index Page Name
- Allow Reports Index page title to no longer be restricted.
- Allow multiple Reports Index Pages under a group

kzadrozny:
Closes #481 

**Changes in this request**
- Removed name requirement of "Reports" from `GroupReportsIndexPage` and `IntranetUnitsReportsIndexPage`
- Removed name requirement of Year from `GroupReportsPage` and `IntranetUnitsReportsPage`

### DIFF
--- a/group/models.py
+++ b/group/models.py
@@ -475,13 +475,6 @@ class GroupReportsIndexPage(BasePage):
     content_panels = Page.content_panels + BasePage.content_panels
     subpage_types = ['group.GroupReportsPage']
 
-    def clean(self):
-        """
-        Make sure page titles adhere to strict
-        formatting policy.
-        """
-        enforce_name_as_reports(self.title)
-
     def get_context(self, request):
         """
         Get reports from children.
@@ -519,13 +512,6 @@ class GroupReportsPage(BasePage):
     ]
 
     subpage_types = ['base.IntranetPlainPage']
-
-    def clean(self):
-        """
-        Make sure page titles adhere to strict
-        formatting policy.
-        """
-        enforce_name_as_year(self.title)
 
     def get_context(self, request):
         """
--- a/group/models.py
+++ b/group/models.py
@@ -67,16 +67,6 @@ def enforce_name_as_year(title):
         )
 
 
-def enforce_name_as_reports(title):
-    """
-    Helper function for making sure
-    page titles adhere to a strict
-    formatting policy.
-    """
-    if not title.strip() == "Reports":
-        raise ValidationError({'title': ('The title should be "Reports"')})
-
-
 def get_page_objects_grouped_by_date(obj):
     """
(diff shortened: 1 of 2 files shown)


intranetunits/models.py
```diff
@@ -38,13 +38,6 @@ class IntranetUnitsReportsIndexPage(BasePage):
     content_panels = Page.content_panels + BasePage.content_panels
     subpage_types = ['intranetunits.IntranetUnitsReportsPage']
 
-    def clean(self):
-        """
-        Make sure page titles adhere to strict
-        formatting policy.
-        """
-        enforce_name_as_reports(self.title)
-
```
bbusenius, line 47: Since both `enforce_name_as_reports` and `enforce_name_as_year` are now unused in this file, we should get rid of them from the import at the top of the page (line 13).

group/models.py
```diff
@@ -520,13 +513,6 @@ class GroupReportsPage(BasePage):
 
     subpage_types = ['base.IntranetPlainPage']
 
-    def clean(self):
-        """
-        Make sure page titles adhere to strict
-        formatting policy.
-        """
-        enforce_name_as_year(self.title)
-
```
bbusenius, line 529: As I can tell both the `enforce_name_as_year` and `enforce_name_as_reports` are no longer used anywhere, so these can be removed from this file. That means deleting lines 53-77.

intranetunits/models.py
```diff
@@ -71,19 +65,14 @@ class IntranetUnitsReportsPage(BasePage):
     ] + BasePage.content_panels
 
     search_fields = BasePage.search_fields + [
-        index.SearchField('get_report_summaries_for_indexing', partial_match=True),
-        index.SearchField('get_report_doc_links_for_indexing', partial_match=True),
+        index.
+        SearchField('get_report_summaries_for_indexing', partial_match=True),
+        index.
+        SearchField('get_report_doc_links_for_indexing', partial_match=True),
```
bbusenius, line 71: I was wrong. My autofixer does this too now! It's annoying but I guess we have to live with it. No big deal.

bbusenius (changes requested):

group/models.py
kzadrozny: `enforce_name_as_year` is still being used by the `GroupMeetingMinutesPage` class as I thought it made sense to retain that convention.

intranetunits/models.py
kzadrozny: Looks like `enforce_name_as_reports` is being used by `GroupMemberRole` still. I am unsure what that class is using the reports function for.

bbusenius (approved):
This looks good. Thanks for making those little changes.

PR #487, opened 2021-03-29
Add permissions check to staff view

### LINKED ISSUE #486: The loop staff page is accessible from the public view
## Expected behavior
User should be redirected to the "you don't have permission to view Loop" page when trying to view `www.lib.uchicago.edu/staff` publicly, if they haven't authenticated over Shibboleth.

## Actual behavior
Loop staff page at [www.lib.uchicago.edu/staff](https://www.lib.uchicago.edu/staff) is open for browsing by the public.

## Other information
Can be fixed by performing the same check we perform in `staff/views.py` to make sure the user is authenticated before viewing Loop search results.  Similar to how we resolved #391.

bufordrat:
Fixes #486.

**Changes in this request**
Add a simple check to staff/views.py to make sure user is authenticated before trying to view Loop staff page.  Follows same approach we used to deal with #391.

### DIFF
--- a/staff/views.py
+++ b/staff/views.py
@@ -1,24 +1,25 @@
+import requests
+
 from base.wagtail_hooks import (
     get_required_groups, has_permission, redirect_users_without_permissions
 )
-from django.db.models import Q
-from django.shortcuts import render
 from django.core.exceptions import PermissionDenied
 from django.http import JsonResponse
-from intranetunits.models import IntranetUnitsIndexPage, IntranetUnitsPage
+from django.shortcuts import render
+from intranetunits.models import IntranetUnitsPage
 from public.models import LocationPage
-from subjects.models import Subject
-from staff.models import StaffPage, StaffPageSubjectPlacement
-from units.views import get_staff_pages_for_unit
-from wagtail.core.models import Page, Site
-from wagtail.images.models import Image
-from wagtail.search.models import Query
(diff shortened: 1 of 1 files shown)


bbusenius (approved):
This is good. It looks like this file was never properly linted before so there's a fair amount of noise no matter what we do. I added a commit just to remove some unused imports and apply the auto-fixers. You'll want to get that change before you merge if you're doing it on the command line.

bufordrat:
> This is good. It looks like this file was never properly linted before so there's a fair amount of noise no matter what we do. I added a commit just to remove some unused imports and apply the auto-fixers. You'll want to get that change before you merge if you're doing it on the command line.

Was my autofixer not working YET AGAIN????

bufordrat:
> > This is good. It looks like this file was never properly linted before so there's a fair amount of noise no matter what we do. I added a commit just to remove some unused imports and apply the auto-fixers. You'll want to get that change before you merge if you're doing it on the command line.
> 
> Was my autofixer not working YET AGAIN????

I swear, it breaks every ten minutes!

bbusenius:
I should have mentioned this on your last pull request. It looks like it is applying `autopep8` just fine but it's missing `isort` and `yapf`. That means your linter is fixing all the Pep8 errors but it's ignoring the code style suggestions.

PR #493, opened 2021-04-01
CGIMail ID not mapped correctly

### LINKED ISSUE #490: CGIMail ID not mapped correctly
**Problem**
- CGIMail to `id` is mapping to `name`
- Produces identical `ids` on one page when radio boxes or check boxes are used (ADA error)
- Changing link 147 to `const id = elm.id || elm.name || null;` did not resolve the issue.

**Example**
`About_this_material` in the radio boxes on this page: https://www.lib.uchicago.edu/search/forms/ask-scrc-or-request-scan/

bbusenius:
Fixes #490

**Changes in this request**
- Proper mapping to IDs for form elements. Will still fallback to names if IDs are not present.


kzadrozny (approved):
Radio boxes acting as expected with unique `id`s - look good!

PR #494, opened 2021-04-01
Added help panel with direct link to Loop staff page

### LINKED ISSUE #491: Staff page editing confusion
Staff are having trouble choosing the correct staff page to edit. Consider utilizing `HelpPanel` to in Staff Public Page admin view to direct staff to Loop for editing. [HelpPanel info](https://docs.wagtail.io/en/v2.12.3/reference/pages/panels.html#wagtail.admin.edit_handlers.HelpPanel)

- Ideal Solution: Link to edit view of staff's Loop page using CNET link
- Medium Solution: Link to Loop staff page with instructions on how to get to edit view
- Base Solution: Link to the Loop staff admin view directory to get them in the right area

Mock:
Screen Shot 2021-03-30 at 4 41 51 PM

kzadrozny:
Closes #491

**Changes in this request**
- Decided to do a `def` only in `StaffPublicPage`
- Note on using `.instance`
   - The base EditHandler class has access to the instance as long as the data is in the model.
   - Value would not populate without `.instance` in the Wagtail admin although functions just fine if calling into a public-facing template.

### DIFF
--- a/public/models.py
+++ b/public/models.py
@@ -13,8 +13,8 @@ from staff.utils import libcal_id_by_email
 from subjects.utils import get_subjects_html
 from units.models import BUILDINGS
 from wagtail.admin.edit_handlers import (
-    FieldPanel, FieldRowPanel, InlinePanel, MultiFieldPanel, ObjectList,
-    PageChooserPanel, StreamFieldPanel, TabbedInterface
+    FieldPanel, FieldRowPanel, HelpPanel, InlinePanel, MultiFieldPanel,
+    ObjectList, PageChooserPanel, StreamFieldPanel, TabbedInterface
 )
 from wagtail.api import APIField
 from wagtail.core import blocks
@@ -843,8 +843,24 @@ class StaffPublicPage(PublicBasePage):
 
     subpage_types = ['public.StandardPage']
     content_panels = Page.content_panels + [
+        HelpPanel(
+            heading='Editing your staff page',
+            template='public/blocks/staffpage_helppanel.html',
+        ),
         FieldPanel('cnetid')
     ] + PublicBasePage.content_panels
+
+    def get_staff_page_id(self):
+        """
+        Gets the page ID from a loop staff page.
+
+        Returns:
+            ID or empty string.
+        """
+        try:
+            return StaffPage.objects.all().filter(cnetid=self.cnetid)[0].id
+        except (IndexError):
+            return ''
 
     def get_bio(self):
         """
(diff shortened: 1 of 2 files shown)


public/models.py
```diff
@@ -843,9 +843,25 @@ class StaffPublicPage(PublicBasePage):
...
+        """
+        Gets the page ID from a loop staff page.
+
+        Returns:
+            ID or empty string.
+        """
+        try:
+            return StaffPage.objects.all().filter(cnetid=self.cnetid)[0].id
+        except (IndexError):
+            return ''
```
bbusenius, line 863: Good use of `try`. I thought we might need this.

public/templates/public/blocks/staffpage_helppanel.html
```diff
@@ -0,0 +1,11 @@
+<fieldset>
+    {% if self.heading %}
+        <legend>{{ self.heading }}</legend>
+    {% endif %}
+    <div class="{{ self.classname }}">
+    	<h2>Want to edit your staff page?</h2>
+    	<h3>Edit your info on your <a href="/admin/pages/{{ self.instance.get_staff_page_id }}/edit/">Loop staff page</a></h3>
```
bbusenius, line 7: This makes sense. Even simpler than what I was thinking. I guess I was thinking we needed to use the `Site` object to get the first part of the url for Loop. But when I see this and think about it more, I can see that the user is already going to be authenticated no matter what and it doesn't matter what URL they came in on :+1:

bbusenius: Interesting to see `self.instance...`. I don't believe we do that anywhere else in the site and I didn't know we could. It makes sense though, especially since we're in the admin view.

bbusenius (approved):
When I look at this, I can see that we're doing `StaffPage.objects.get(cnetid=self.cnetid)` and `StaffPage.objects.all().filter(cnetid=self.cnetid)` in a few different places. I feel like this could be more efficient if we only did it once for the page object and then used the same object in the different places. That being said, I think that might be more fussing than you're wanting to do with this and I'm pretty sure this won't slow things down since it's a function and only called where it is being used in the admin template. I'm not asking you to make that change. That might be the kind of restructuring we would want to do in the future (especially if we ran into performance issues) or something to think about in the future if we're feeling more ambitious with a particular piece of code. I'm willing to bet we have that problem in other places too. It's a pattern we use frequently. Anyhow, that's just food for thought. This looks good!

kzadrozny:
Created an issue for making the StaffPage functions more efficient in #495

PR #498, opened 2021-04-08
479 retain search term when switching source

### LINKED ISSUE #479: Retain search term when switching source
Librarian request: "On the ‘Articles, Journals & Databases’ search tab on the kiosk pages, it would be possible to retain one’s search terms when changing the radio button."

Will use VuFind's js as a base to retain terms between switching search box sources / views.

bufordrat:
Fixes #479.

Add a little javascript to `transformer-tabs.js`, plus a couple extra `id`-s and `onchange` events to `search_widget.html`, to get the following effect:
- when the user types something in any of the search boxes on the main page, that something appears in all the search boxes
- when the user erases what's in any of the search boxes on the main page, they all go back to normal

### DIFF
--- a/base/static/base/js/transformer-tabs.js
+++ b/base/static/base/js/transformer-tabs.js
@@ -56,3 +56,23 @@ var Tabs = {
 }
 
 Tabs.init();
+
+// propagate an input element's content to all the other forms in the
+// search box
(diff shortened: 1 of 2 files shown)


bufordrat:
Thanks!  Glad you enjoyed it.

PR #504, opened 2021-04-14
Changed list-box location

### LINKED ISSUE #499: SiteImprove Carousel Restructure
## Problem
- `list-item` is being used without a list
- Carousel items are in nested `div`s and not up to W3C's standards
- Example page: https://www.lib.uchicago.edu/mansueto/

## Resolve
- Restructure carousel block into `ul` `li` and `article`
- W3C's recommended structure: https://www.w3.org/WAI/tutorials/carousels/structure/

kzadrozny:
Closes #499

**Notes**
Difficult to shoehorn the ul / li structure into the Bootstrap carousel framework:
- Opted to move `listbox` to the already exisiting `ol`
- Swapped carousel item from `div` to `article` to fit typical ADA structure
- Need to revisit if we want to move away from a Boostrap option to something else. For now, resolves the SiteImprove error.

### DIFF
--- a/base/templates/base/public_base.html
+++ b/base/templates/base/public_base.html
@@ -193,18 +193,18 @@
                     <div id="carousel-example-generic" class="carousel slide" data-ride="carousel" data-interval="false">
 
                         {% if carousel_multi %} 
-                            <ol class="carousel-indicators">
+                            <ol class="carousel-indicators" role="listbox">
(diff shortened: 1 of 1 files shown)


bufordrat (approved):
Looks great!  Tried it out on `nest` and at least to my naked eye looks just like the old carousel.  And the html resembles what the issues says it should look like.

PR #527, opened 2021-06-22
Fix unit tests for upgrade

### LINKED ISSUE #526: Fix unit tests for upgrade
## Expected behavior
Unit tests should always pass. This is important prior to upgrading since we need to use the tests to test upgrades.

## Actual behavior
Unit tests are failing.

### Steps to reproduce
`./manage.py test`

bbusenius:
Fixes #526

**Changes in this request**
- Pegged `psycopg2` and `django-webpack-loader` to version numbers.
- Updated the html of a test to match what is in the template. The template changed as part of the ADA updates and the test needed to be altered to reflect that.

### DIFF
--- a/lib_collections/tests.py
+++ b/lib_collections/tests.py
@@ -59,7 +59,7 @@ class test_lib_collections_view(TestCase):
         response = collections(request)
 
         self.assertContains(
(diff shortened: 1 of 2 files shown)


kzadrozny (approved):
Nice catch on the aria-label!

PR #535, opened 2021-06-24
Forgotten migration

### LINKED ISSUE #534: Forgotten migration
## Behavior
In working on the problem with the upgrade, I discovered that we are missing a migration. This is something related to the changing of building names. I'm a little surprised this triggered a migration. Anyhow, it did and we need it.

*Note: This does _not_ trigger a change in fixtures when I `dumpdata`. That's why fixtures are not included in the PR.

bbusenius:
Fixes #534

Doing this as a separate PR in order to keep the upgrade branch clean.

**Changes in this request**
- Adds a missing migration.

*Note: This does not trigger a change in fixtures when I `dumpdata`. That's why fixtures are `not` included in the PR.


kzadrozny (approved):

bufordrat (approved):
Works on my Vagrant instance; can't find any pages acting weird after running `manage.py migrate`.

PR #543, opened 2021-07-26
Added reusable content to Location Page widgets

### LINKED ISSUE #540: Reusable Content Box on Floorplan Template
Request to include the reusable content box on floorplan pages to allow for mass messaging about building closures. Since floorplans do not have an editable sidebar, this will likely be placed into the content of the page, above the map image. Needed for Aug 2nd, ideally.

kzadrozny:
Fixes #540
Note: Mis-typed the original request -- the addition is for LOCATION pages, not floorplan pages.

**Changes in this request**
- Added a `page alert` streamfield section in the widgets tab
- Allows for paragraph, HTML, or reusable content
- Included these options for flexibility (unique in-page alert or ability to add LibGuides widget)
- Did light CSS clean-up for banner alert (above nav) and features list in Location sidebar


bbusenius (approved):
This looks totally fine, I don't see any problems in the code. Thank you for including fixtures.

I was a little surprised when I saw this because it is so different from how we're already using reusable content but then I looked at the code and saw that 1. these pages were not inheriting the `reusable_content` field and 2. they do not have the default body StreamFields. So, basically everything I said in our earlier meeting was wrong. I guess that's what I get for talking about code 100% from memory. Anyhow, it seems like this use case is very specific, since it's being called `page_alerts` and only displays in 1 place, unlike reusable content that is more multipurposable in StandardPages where it can be in the sidebar or body. Anyhow, I don't have a problem with any of that. I'm just talking out loud. Since the LocationPages are unique, I think this makes sense.

kzadrozny:
I agree - I was hesitant to use this approach, but I saw we were doing something similar in the Ask a Librarian pages. Since the Location pages are so specific, I think (and hope) it will work for the needs presented.

PR #547, opened 2021-08-12
Added for loop directly in Reports Index template

### LINKED ISSUE #546: Report Index Page Error
**Problem**
When there is more than one Report Index Page at the same level both Index pages lists all the report pages under them. This leads to the mixing of all reports pages that are children of the Index page and any others at that same sublevel.

**Solution**
Needs to tweak the Index Page model so it only lists its direct child pages

kzadrozny:
Fixes #546 .

**Changes in this request**
- Moved for loop with dictionary list into the Report Page Index
- The structure was calling both child pages and uploaded links/docs specified from "Year" page. Moving `dl` to index seems to skip over this erroneous call. 
- Reviewed by EE, who reported the issue, so verify concern was resolved.

Screen Shot 2021-08-12 at 1 01 14 PM
Screen Shot 2021-08-12 at 1 01 18 PM

### DIFF
--- a/group/templates/group/group_reports_page.html
+++ b/group/templates/group/group_reports_page.html
@@ -7,7 +7,16 @@
 	<article>
         <h1>{{ self.title }}</h1>
         {% with table=self.get_reports_grouped_by_date %}
-            {% include "base/minutes_and_reports_outer_table.html" %}
+            <dl>
+                {% for date, items in table.items %}
+                  <dt>{{date}}</dt>
+                  {% for item in items %}
+                    <dd>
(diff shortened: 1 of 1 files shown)


bbusenius (approved):
This looks fine. I see we cannot delete the `minutes_and_reports_outer_table.html` template because it is being used elsewhere.

kzadrozny:
I plan to delete both `minutes_and_reports_outer_table.html` and `minutes_and_reports_inner_table.html` when I rework the Meeting Minutes.

PR #554, opened 2021-09-03
get_token and libcal_id_by_email now catch all request exceptions

### LINKED ISSUE #553: LibCal Button Not Handling Request Exceptions
## Expected behavior
LibCal button on staff page and on front page should fail elegantly in the event that `https://rooms.lib.uchicago.edu` is not available.

## Actual behavior
When running the Wagtail site from a machine with a local IP but not WAN IP, a request exception gets thrown, because LibCal has nowhere to send a response to.

### Steps to reproduce
This is tricky.  Contact @bbusenius for more information!

## Fix
Can be fixed by making sure to have LibCal button catch all possible exceptions to do with requests, like timeouts or refused connections.  The logic for this is in `staff/utils.py`: specifically in `get_token` and `lookup_staff_ids`.

bufordrat:
Fixes #553.

**Changes in this request**

Uses `requests.exceptions.RequestException` to catch a broad class of exceptions due to timeouts, refused connections, etc.
- 
- 
-

### DIFF
--- a/staff/utils.py
+++ b/staff/utils.py
@@ -781,7 +781,7 @@ def get_token(url, data):
     try:
         resp = requests.post(url, data)
         return resp.json()['access_token']
-    except(KeyError, requests.exceptions.MissingSchema):
+    except(KeyError, requests.exceptions.RequestException):
         return ''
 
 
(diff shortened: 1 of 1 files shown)


bbusenius (approved):
This looks good to me. Simple fix to make things safer. Thanks for taking the time to look at this.

PR #556, opened 2021-09-13
Item Servlet

### LINKED ISSUE #555: Item Servlet
## Need
We need to recreate a Folio driven version of the item servlet that allows us to get bib and item information for records and plug it into a webpage. The primary use case for this is to pre-populate CGIMail forms on the website, however, it can be used anywhere were we want to drive content using bib numbers and barcodes from the catalog. We currently have a version of this written in PHP and driven by OLE that exists outside of the Library website. This new version will be part of the website like our other APIs.

## Related materials
[Current item servlet](http://forms2.lib.uchicago.edu/lib/searchform/itemServlet.php?format=json&bib=3&barcode=18369245&database=testing) (This is the existing version driven by OLE).

bbusenius:
Fixes #555 

To see this work, go to `/item-servlet/?bib=9291257&barcode=109010396`. You can pass in bib numbers and barcodes. These should be from the same record in the catalog. 

**Changes in this request**
- Implementation of a new item servlet API driven by Folio to replace the OLE one.


bufordrat (approved):
This is awesome!  I really like the way this code takes in info the way Folio gives it to you and then massages it into the form we're going to use.  And I like the way it handles exceptions.  Totally worked with the example URL once I got my Vagrant VM up and running.

PR #578, opened 2021-12-13
574 index features tags from library spaces

### LINKED ISSUE #574: Index Features tags from Library Spaces pages
Please add the Features tags to the indexing of page content for public search. The features are selected by checkbox in the staff interface, and include "Quiet Zone", "Standing Desk", "No Food", etc. The desired outcome is that if a user searches for "Standing Desk", search results would include all pages with this tag.

bufordrat:
Fixes #574.

**Changes in this request**
Adds a tiny amount of code to `public/models.py` and `public/utils.py` to throw the spaces locations into the search fields for `LocationPage`-s.

### DIFF
--- a/public/models.py
+++ b/public/models.py
@@ -699,7 +699,6 @@ class LocationPage(PublicBasePage, Email, Address, PhoneNumber):
 
     subpage_types = ['public.StandardPage', 'public.FloorPlanPage']
 
-    # location_strings = post_process([item[0] for item in get_features()])
     location_fields = [mk_search_field(item[0]) for item in get_features()]
 
     search_fields = PublicBasePage.search_fields + [
--- a/public/models.py
+++ b/public/models.py
@@ -8,7 +8,7 @@ from base.models import (
 from django.db import models
 from django.db.models.fields import CharField
 from modelcluster.fields import ParentalKey
-from public.utils import get_features
+from public.utils import get_features, mk_search_field
 from staff.models import StaffPage
 from staff.utils import libcal_id_by_email
 from subjects.utils import get_subjects_html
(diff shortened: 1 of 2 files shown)


bbusenius (approved):
This looks fine. Another way to have accomplished this would have been to have `mk_search_field` return a string that joined all of the values from `get_features` and then just have indexed that (would have needed to be a method in the page model and not in utils), e.g.

```
def mk_search_field(self):
    return ' '.join(i[0] for i in get_features())

   search_fields = PublicBasePage.search_fields + [
        index.SearchField('short_description', partial_match=True),
        index.SearchField('long_description', partial_match=True),
        index.SearchField('parent_building'),
        index.SearchField('location_photo'),
        index.SearchField('reservation_url'),
        index.SearchField('reservation_display_text'),
        index.SearchField('mk_search_field'),
    ]
```

bufordrat:
> ```
> def mk_search_field(self):
>     return ' '.join(i[0] for i in get_features())
> ```

Very interesting.  So just to make sure I'm understanding this alternate method, the search backend apparatus would understand how to turn a string like `"has_all_gender_restrooms has_lockers has_large_tables"` into what it needs for the search index?

bufordrat:
Or rather, let me rephrase that.  The best way I can make sense of that is that: Wagtail's `index.SearchField` class takes in a string which has to be the name of the method for the type of page whose fields you're adding to the index.  It looks like the method can return whatever--in the case of the individual field names it's a boolean and in the case of your version of `mk_search_field` it's a big string, and then what `index.SearchField` does is `eval` that method name, adding both the method name itself and whatever `self.the_method_name()` evaluates to to the search index, in a way that magically ignores underscores, spaces, capitalization, etc.

bufordrat:
Update: I tried deploying my old code and it didn't work!  Only this code did.  Looks like all is well now.

bbusenius:
To answer your question, normally Wagtail just takes a field name and indexes the field. Providing a method is the alternative way that is also supported. It allows you to index things that aren't fields (or other more complex scenarios). The method can be named whatever you want.

PR #586, opened 2022-02-28
Item servlet fails in some cases

### LINKED ISSUE #585: Item servlet fails in some cases
## Expected behavior
The item servlet should return holdings information when item information is not found. Furthermore, item information should always be found when items exist. 

## Actual behavior
Currently holdings data fails to return when items are not found. Additionally, items are being scoped to the holdings ID, which normally works, however, it fails in these instances. This can be resolved by using barcode to make the query rather than using it to filter the query.

### Steps to reproduce
[Go to this record](https://catalog.lib.uchicago.edu/vufind/Record/1299992) and click on the "Request via Interlibrary Loan". It will fail. You may also compare this by going to the item servlet directly with the query: https://www.lib.uchicago.edu/item-servlet/?p=true&bib=1299992&barcode=088440577. This will also fail.

## Testing notes
http://wwwdev:8000/item-servlet/?p=true&bib=1299992&barcode=088440577 should populate with item information in the fixed branch. If it does not, empty the Wagtail cache in the admin.

bbusenius:
Fixes #585 

**Changes in this request**
- Allow bib level data to populate if items data is not present (was failing in select cases)
- Get items by barcode instead of filtering holdings > items by barcode.

**Testing Notes**:
- Fail case: https://catalog.lib.uchicago.edu/vufind/Record/1299992
- http://wwwdev:8000/item-servlet/?p=true&bib=1299992&barcode=088440577 should populate with item information, if it does not, empty the Wagtail cache in the admin.

### DIFF
--- a/item_servlet/tests.py
+++ b/item_servlet/tests.py
@@ -69,7 +69,7 @@ class TestItemServletUtils(TestCase):
         self.assertTrue(holdings == {})
 
     def test_get_item_with_bad_data(self):
-        item = get_item({'foo': 'bar'}, 'badbarcode', 'badtoken')
+        item = get_item('badbarcode', 'badtoken')
         self.assertTrue(item == {})
 
     def test_get_location_with_bad_data(self):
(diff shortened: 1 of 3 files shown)


bufordrat (approved):
Ooh, yes!  Makes perfect sense to get the holdings separately and then not pass them again into `get_item`.  I like the small and simple fixes in all the right places.  And all the examples I tried work in dev mode on my machine.

I guess it previously seemed like we needed the holdings for the FOLIO Okapi API, but now we know we don't.

PR #614, opened 2022-09-15
Additional security patches

bbusenius:
More `npm` updates.


bbusenius:
For some reason these updates output identical static js files for us. I think that makes this really safe because it means the generated javascript is identical. I'm guessing this is because these are dependencies of the dependencies as well as dev dependencies.

bluairis (approved):
- Vagrant works on windows and mac
- Forms submit (only tested on mac)

bufordrat (approved):
Vagrant VM builds for me too!

Issue #628, opened 2022-12-01
LibAnswers phasing out old domain names


bbusenius:
## Problem
Springshare is phasing out older domain names for LibAnswer APIs and widgets. We need to identify and update old API calls and widgets to make sure we are using the proper syntax.

### Examples of URL domains that are being phased out:
- v2.libanswers.com
- region-xx.libanswers.com
- api2.libanswers.com
- api2-xx.libanswers.com

### The above old URL domains should be replaced with:
- uchicago.libanswers.com

_Only the base part of the URL should be replaced._

#### Code changes
Most of the changes that are needed are found in Wagtail pages or widgets we've used. These must be found and updated manually. The Library chat status functionality is using the correct domain which it pulls from a config file. The only code change that is needed are old domains in the `AskPage` templates.
- Update `AskPage` templates.

## Related materials
- [Spreadsheet identifying places where we might need to update urls](https://docs.google.com/spreadsheets/d/1KVDky52KLQj5j0w30ZaYTstEgm_Vjko_2WpH58ZCD_U/edit#gid=0).

PR #629, opened 2022-12-01
LibAnswers phasing out old domain names

bbusenius:
Related to issue #628 

**Changes in this request**
- Update LibAnswers domain to use the new syntax.
- Offloaded more to the config file and removed repetitive code.


bufordrat (approved):
Looks great!

PR #632, opened 2022-12-05
Need ability to suppress staff from directory sync command

### LINKED ISSUE #630: Need ability to suppress staff from directory sync command
## Problem
Sometimes a staff member in the Library directory needs to be out of sync with the University directory. In rare cases the staff member will not even exist in the University directory. We need a way to suppress staff members so that they don't trigger an "out of sync" email report every day when this occurs.

bbusenius:
Fixes #630 
**Changes in this request**
- Made the suppression of staff from the out of sync report configurable.

_Note: I tried to write a unit test for this but found that 1. this part of the code is very difficult to test (even after some refactoring) and 2. it adds too much time to the unit tests (more than 1 minute). We will need to rewrite significant portions of this part of the code to make it more testable._


bufordrat (approved):
Excellent!  Love the idea of cleaning these email notifications up.  Tried removing someone from the list on `nest` and wouldn't you know it, they then appear in the report.

PR #634, opened 2022-12-09
completed issue on correct version of wagtail

### LINKED ISSUE #595: Add function to override header font size and kerning in Collections and Exhibits
SCRC staff would like advanced controls for header text in Collections and Exhibits to allow for fonts to be resized or spacing to be adjusted.

bluairis:
Fixes #595  

- These are the same changes that were made in the previous #595 pull request except that they are on the new version of wagtail and I have also added the migrations to the commit


bbusenius (approved):
This looks good to me! The code changes are the same as on the other branch. I tested it and the functionality works as intended. It also passes unit tests. I don't think we need `fixtures` in this case because we're not preserving any data in the dev database. My only recommendation for next time is that you resolve conflicts in git rather than open a new branch.

bufordrat:
> My only recommendation for next time is that you resolve conflicts in git rather than open a new branch.

Oh, just to clarify about that, creating a new branch was my suggestion because the old code was on the < 4 version of Wagtail.

bufordrat (approved):
Excellent!  I was kerning away.  For future reference: if we want, we can add a validation on the font scaler so that it rejects negative numbers.  Here is a guide on how to do that:
https://stackoverflow.com/questions/54858123/how-do-i-enforce-a-positive-number-on-my-python-django-model-field

PR #636, opened 2022-12-10
616 make loop and regular redirect pages identical

### LINKED ISSUE #616: Make Loop and Regular Redirect Pages Identical
- The loop redirect pages do not have a link document and the link internal page might be broken, but since:
- They also require a link, I can't test the link internal page
- They also have a default

bluairis:
Fixes #616 

**Changes in this request**
- Made loop redirect pages look like regular redirect pages
- Added a document field to them
- Not having a default link
- Not having anything required


bbusenius (approved):
This looks good. It works as expected and passes tests.

bufordrat (commented):
Noted for future reference: if you populate all three of the redirect links, they overwrite each other, apparently in this (descending) priority order:

1) internal link page
2) document link
3) external link

Not clear why without looking at the code further, but that's something we can look into.  Anyway, the PR looks good, just thought I'd leave this note here in case we need to refer to it later.

bufordrat (approved):
Tested this on `nest` and everything looks great.  We ran into weirdness choosing a unit to go with the standard `www.lib` redirect page, but that is a separate issue we'll have to look into at another time.

PR #640, opened 2023-01-04
404 Pages for Loop

### LINKED ISSUE #594: 404 page for Loop
EE
I just hit a funny error. This page obviously doesn't exist: https://loop.lib.uchicago.edu/departments/administration/communications/lib/
But the main nav is active on this page, but since I got there from Loop, it tries to take me to a Loop-addressed equivalent: https://loop.lib.uchicago.edu/crerar/
Which also obviously doesn't exist.
How do we fix this so that Loop has its own 404 (or whatever is required so that Loop address errors aren't kicked out into a weird simulacrum of the main website)?

FS
I didn’t understand the issue here so I had to go ask.  The problem is that the the custom ‘404’ error page that loop.lib is displaying is actually from the www.lib server.  In other words, even though the URL shows loop.lib, all of the content (and all of the menus) are from from www.lib.  This means that if the user clicks on any of the links on this page, 90% of them also generate a 404 error (because they don’t exist on loop.lib).

Best solution - loop.lib should have its own 404 page, separate from the 404 page that www.lib uses.

bluairis:
Fixes #594 
**Changes in this request**

- Created a template that talks to the views to determine if a loop or library 404 page should be posted
- Koala image on loop page on the right of the screen
- Loop has a toggle search while the regular 404 does not
- Library has bullets for additional helpful links, but loop does not because there is only 1


library_website/templates/404.html
```diff
@@ -1,21 +1,144 @@
...
 {% block body_class %}template-404{% endblock %}
 
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.4.1/css/bootstrap.min.css">
+  <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.6.1/jquery.min.js"></script>
+  <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.4.1/js/bootstrap.min.js"></script>
+</head>
```
bufordrat, line 13: @bbusenius isn't there another place in the Wagtail site where we usually stick jQuery, Bootstrap, etc. or am I misremembering?  One of the includes in the `base` app, maybe?

library_website/templates/404.html
```diff
@@ -1,21 +1,144 @@
...
     <h1>Page not found</h1>
-    <p>You may have an outdated bookmark or we have a broken link.</p>
 
+
+    {% if koala %}
+    <div class="block-image">
+    	<figure class="imgcaption pull-right">
+    	{% image koala width-200 %}
+	</figure>
+    </div>
```
bufordrat, line 28: NICE

library_website/templates/404.html
```diff
@@ -1,21 +1,144 @@
...
+      {% endif %}
+	
+
+    <h2>Try a Sitewide Search</h2>
+
+
+    {% if tabs %}
+   
+    	<ul class="nav nav-pills">
+
```
bufordrat, line 64: You could maybe trim some of these additional line breaks.

library_website/templates/404.html
```diff
@@ -1,21 +1,144 @@
...
+	  <li class="active"><a data-toggle="pill" href="#loop_website">Loop</a></li>
+
+	  <li><a data-toggle="pill" href="#library_website">Library Website</a></li>
+
+	</ul>
+
+
```
bufordrat, line 71: ditto

library_website/templates/404.html
```diff
@@ -1,21 +1,144 @@
...
+	 
+
+	  <div id="loop_website" class="tab-pane fade in active">
+
+	    <br>
+
+
+	    <form action="{% url 'search' %}" method="">
+	      <div class="col-xs-9 col-sm-10 nopadding">
+		<input name="query" type="text" placeholder="search loop" aria-label="..." class="form-control" id="search_form5" onchange="propagate(this)">
```
bufordrat, line 82: I believe you can get rid of `onchange=propagate(this)`, since as I recall that is for the actual search bar on the main page.  (As in: the `propagate` Javascript function is exposed there but I don't think this code can see it.)

library_website/templates/404.html
```diff
@@ -1,21 +1,144 @@
...
+                             
+	      <div class="col-xs-3 col-sm-2 search-btn-wrapper">
+		<button class="btn btn-search btn-default" type="submit"><span class="hidden-md hidden-lg glyphicon glyphicon-search"></span><span class="hidden-xs hidden-sm">Search</span></button>
+	      </div>
+	    </form>
+
+	  </div>
+	 
+
+	  <div id="library_website" class="tab-pane fade">
```
bufordrat, line 95: Pretty!

library_website/templates/404.html
```diff
@@ -1,21 +1,144 @@
...
+
+
+	    <br>
+
+
+	    <form action="/switchboard/" method="post">
+	      <div class="col-xs-9 col-sm-10 nopadding">
+		<input name="query" type="text" placeholder="search library website" aria-label="..." class="form-control" id="search_form5" onchange="propagate(this)">
```
bufordrat, line 103: ditto for all the rest of these examples of `onchange=propagate(this)`...

library_website/views.py
```diff
@@ -0,0 +1,62 @@
...
+                ],
+                "helptext1":
+                "Let us know that we have a missing page. Contact ",
+                "helptext_email":
+                "intranet@lib.uchicago.edu",
+                "helptext2":
+                " and we will do our best to rectify the issue.",
+                "tabs":
+                True
+            }
```
bufordrat, line 58: Great separation of content from layout here!

bufordrat (changes requested):
This looks great; I just had a couple small suggested changes.

library_website/templates/404.html
bbusenius: Yes, normally this is inherited from the respective base templates. This was something that struck me too. I'm guessing the reason for it is that the 404 templates sit outside of the infrastructure (they're not Wagtail page objects)? I was going to look at this closer tomorrow.

bbusenius: Actually, I think all of the wrapper html is extraneous and unnecessary. That includes the `html`, `head`, and `body` tags. From what I can tell, most of this isn't being included in the rendered html, except for one of the `body` close tags (which means there are two and it's producing invalid html). That's because the template won't use anything that isn't inside one of the blocks: `{% block content %}` and `{% block body_class %}` in this case. So, everything that's outside of `{% block content %}` and `{% block body_class %}`  should be eliminated. More on template inheritance here: https://docs.djangoproject.com/en/4.1/ref/templates/language/#template-inheritance

library_website/templates/404.html
```diff
@@ -1,21 +1,144 @@
...
 {% block content %}
 <div class="col-xs-12 centermain" style="padding-top:2em;">
+
+   
     <h1>Page not found</h1>
-    <p>You may have an outdated bookmark or we have a broken link.</p>
 
+
+    {% if koala %}
+    <div class="block-image">
```
bbusenius, line 24: This should be indented.

library_website/templates/404.html
```diff
@@ -1,21 +1,144 @@
...
+    	<figure class="imgcaption pull-right">
+    	{% image koala width-200 %}
```
bbusenius, line 26: This should be indented too.

library_website/templates/404.html
```diff
@@ -1,21 +1,144 @@
...
+	</figure>
+    </div>
+    {% endif %}
+    <div class="block-pagraph">
+    <p>You may have an outdated bookmark or we have a broken link.</p>
```
bbusenius, line 31: This should be indented.

library_website/templates/404.html
```diff
@@ -1,21 +1,144 @@
...
+	    </form>
+
+
+	  </div>
+
+     {% else %}
+
+	    <form action="/switchboard/" method="post">
+	      <div class="col-xs-9 col-sm-10 nopadding">
+		<input name="query" type="text" placeholder="search library website" aria-label="..." class="form-control" id="search_form5" onchange="propagate(this)">
```
bbusenius, line 120: This doesn't look like a valid `aria-label` to me. Maybe there's something to this I don't know about but if not, I would get rid of it.

library_website/views.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def context_404(site_name):
+        if site_name == "Public":
+            return {
+                "extention":
+                "base/public_base.html",
+                "links": [
+                    Link(
+                        "Ask a Librarian",
+                        "https://www.lib.uchicago.edu/research/help/ask-librarian/"
```
bbusenius, line 25: This should be a site relative link, not absolute. That way it will play nice in our dev environments and on `nest`.

library_website/views.py
```diff
@@ -0,0 +1,62 @@
...
+                    ),
+                    Link(
+                        "Library Digital Collections",
+                        "https://www.lib.uchicago.edu/collex/"
```
bbusenius, line 29: Site relative link, not absolute.

library_website/views.py
```diff
@@ -0,0 +1,62 @@
...
+                    ),
+                    Link("Library Homepage", "https://www.lib.uchicago.edu/")
```
bbusenius, line 31: Site relative link, not absolute.

library_website/views.py
```diff
@@ -0,0 +1,62 @@
...
+        else:
+            return {
+                "extention":
+                "intranethome/intranet_home_page.html",
+                "koala":
+                Image.objects.get(title="Adventure Koala"),
+                "links": [
+                    Link(
+                        "Return to Loop Homepage",
+                        "https://loop.lib.uchicago.edu/"
```
bbusenius, line 47: Site relative link, not absolute.

bbusenius (changes requested):
This is off to a good start! I called out some issues in the code. The biggest problem is some extraneous markup that is producing invalid html. I think this documentation about template inheritance should help you understand why that markup isn't needed (and isn't being used for the most part): https://docs.djangoproject.com/en/4.1/ref/templates/language/#template-inheritance

bbusenius:
I may have found another issue with this. In looking at something totally unrelated to this PR, the following happened when I submitted a test CGIMail form on my dev machine:

image

We'll need to figure this out too.

bufordrat (approved):
I can dig it.

bufordrat:
FWIW, I do not seem to be getting the sneak embedded 404 page when I submit a CGI form on `nest`.

bufordrat:
Don't know whether there's some range of CGIMail forms we wanna test that on, but I imagine @bbusenius has views on that topic.

library_website/views.py
```diff
@@ -0,0 +1,47 @@
...
+                "helptext1":
+                "Let us know that we have a missing page.",
+                "helptext2":
+                " We will do our best to rectify the issue."
+            }
+        else:
+            return {
+                "404_page_template_switcher":
+                "intranethome/intranet_home_page.html",
+                "koala": Image.objects.get(title="Adventure Koala"),
```
bbusenius, line 36: This should probably have a safe fallback for when the image isn't present and then in the template it should only call the image tag if the image is present.

bbusenius (changes requested):
This is really close now. Besides the minor issue with the broken link, I see one more problem. This code breaks one of the unit tests (`test_non_existing_sub_url_on_article_page` in `lib_news/tests.py`). At a glance I thought it was because the test was expecting a `404` but now it's getting a `200` or a redirect. This might be part of the problem, I'm not sure, but in looking at the test output closer I can see that it's blowing up because the koala image is missing. To solve this, I would put in some handling so that the code fails gracefully if the koala isn't present. Then if the test still fails, it will just need to be updated to match the new conditions.

Run all tests: `./manage.py test --parallel`
Run specific test: `./manage.py test lib_news.tests.test_lib_news_pages_and_categories.test_non_existing_sub_url_on_article_page`

I'm unable to trigger the error with CGIMail forms in dev again and since @bufordrat can't trigger it on `nest`, I think we're in the clear on that one.

bbusenius (approved):
This looks good! All the tests are passing for me now.

PR #648, opened 2023-02-02
Vagrant builds fail because of openpyxl 

### LINKED ISSUE #647: Vagrant builds fail because of openpyxl
## Problem
New Vagrant builds fail with the following error because the `save_virtual_workbook` function was removed from `openpyxl` and we don't have that dependency pinned to a hard version:

```
ImportError: cannot import name 'save_virtual_workbook' from 'openpyxl.writer.excel' (/home/vagrant/lw/lib/python3.9/site-packages/openpyxl/writer/excel.py)
```

### Steps to reproduce
Run:
`vagrant destroy` and `vagrant up`.

### Related information
- https://stackoverflow.com/questions/75299506/cannot-import-name-save-virtual-workbook-from-openpyxl-writer-excel
- https://foss.heptapod.net/openpyxl/openpyxl/-/blob/3.0.10/openpyxl/writer/excel.py#L297
- https://openpyxl.readthedocs.io/en/stable/tutorial.html?highlight=NamedTemporaryFile#saving-as-a-stream

### Related code
- `staff/management/commands/list_staff_wagtail.py`
- `staff/wagtail_hooks.py`
- `units/management/commands/list_units_wagtail.py`
- `units/views.py`

## Other information:
This can easily be fixed temporarily by hard-pinning the `openpyxl` dependency to version `3.0.10` (the last version where the `save_virtual_workbook` function existed) in `requirements.txt`, however, it would be more prudent to set it as `openpyxl>=3.1.0,<3.2.0` and fix our code using `NamedTemporaryFile` as mentioned in the code where this was deprecated.

bbusenius:
Fixes #647 

**Changes in this request**
- Added a new `save_virtual_workbook` function to `base_utils.py` [(line 470)](https://github.com/uchicago-library/library_website/pull/648/files#diff-7d6137b6e6d905f7f7cd657953ba62d3ac6e351b31b5ea30b56cc70b8122dc28R470) to replace the one deprecated in `openpyxl`.
- Changed imports to use the new function instead of the old one.
- Pinned openpyxl to a semantic version: `openpyxl>=3.1.0,<3.2.0`.
- Linted files with black.

### DIFF
--- a/base/utils.py
+++ b/base/utils.py
@@ -1,23 +1,34 @@
 import base64
 import json
 from http.client import HTTPSConnection
+from tempfile import NamedTemporaryFile
 
 import requests
 from django.utils.text import slugify
 from library_website.settings import (
-    ADDRESS_TEMPLATE, CRERAR_HOMEPAGE, DANGELO_HOMEPAGE, ECKHART_HOMEPAGE,
-    HOURS_PAGE, HOURS_TEMPLATE, LIBCAL_IID, MANSUETO_HOMEPAGE, SCRC_HOMEPAGE,
-    SSA_HOMEPAGE
+    ADDRESS_TEMPLATE,
+    CRERAR_HOMEPAGE,
+    DANGELO_HOMEPAGE,
+    ECKHART_HOMEPAGE,
+    HOURS_PAGE,
+    HOURS_TEMPLATE,
+    LIBCAL_IID,
+    MANSUETO_HOMEPAGE,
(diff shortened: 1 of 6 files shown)


bufordrat (approved):
Excellent.  Linting looks good and the branch seems to work on my machine in dev mode.

bluairis:
The build works! It is really cool that you do not need to do any of the source lw/bin/activate and cd /vagrant things!

bluairis (approved):

PR #656, opened 2023-02-10
649 search bar in lib search results page

### LINKED ISSUE #649: Search bar in lib search results page
As discussed, adding a search bar to the search results page in the library to allow a research.

bluairis:
Fixes #649.

**Changes in this request**

- adds a search bar underneath the search results
- the search bar has the previously searched term preloaded allowing the user to re-search based on previous search
- search bar is small on larger screen sizes

### DIFF
--- a/results/templates/results/results.html
+++ b/results/templates/results/results.html
@@ -56,6 +56,17 @@
         {% elif search_query %}
             No results found
         {% endif %}
+	    <form action="/switchboard/" method="post">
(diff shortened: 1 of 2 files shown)


bufordrat (changes requested):
Irina and I just noticed that "no results" appears right-aligned.  So maybe make it left aligned under the search text field?  Otherwise this looks great.

bufordrat (approved):
New left-alignment looks excellent!

PR #660, opened 2023-02-22
Search inputs are missing accessible text

### LINKED ISSUE #659: Search inputs are missing accessible text
## LIB20230113-06. Search inputs are missing accessible text (Blocker)
**Issue**: There are no labels on the Articles Plus and Website search inputs. These inputs do have aria-label="..." attributes, which do not convey the purpose of the inputs. The aria-labels on each input should be updated to provide meaningful accessible text/labeling for these inputs. 

## Related materials
[Full issue description](https://docs.google.com/document/d/15B0iu5qEQOI8BH-P48pgJul1-8A5b9d6Bi6PwoQvlTc/edit#heading=h.h09tizb0aqae)

bbusenius:
Fixes #659 

**Changes in this request**
- Replaces bad aria labels with meaningful text.
- Lints templates.

To view changes without linting whitespace differences, run this command when on the working branch:
```
git diff -w master
```

### DIFF
--- a/base/static/base/js/transformer-tabs.js
+++ b/base/static/base/js/transformer-tabs.js
@@ -63,7 +63,7 @@ const propagate = (element) => {
 
     // get all the forms on the search box page
     const form1 = document.getElementById("search_form1");
-    const form2 = document.getElementById("search_form2");
+    const form2 = document.getElementById("ebscohostsearchtext");
     const form3 = document.getElementById("search_form3");
     const form4 = document.getElementById("search_form4");
     const form5 = document.getElementById("search_form5");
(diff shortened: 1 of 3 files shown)


bufordrat (approved):
Lovely!  Nice to see some meaningful aria labels in there.

bluairis (commented):
I might be wrong about this, because I am not quite sure why it is happening, but I was trying to test the aria labels by turning the voice over reader on, on my iMac. The interesting thing that I found is that the voice over does read the jourals search field differently when I am on wwwdev versus the current library website, however it doesn't work for the databases search tab or the articles search tab. In fact, it just reads them as search bars and doesn't describe them even though it does describe the journals search bars. Long story short, I am a bit confused why this is happening and I do not know if it is the aria label's fault exactly, but I thought it might be worth looking into.

bbusenius:
@bluairis good catch! I'm not sure if I've fixed it but there's a good chance. I rearranged the html attributes on the inputs to match the one that was working and in doing so I found that the first one had an invalid second `id` attribute. That was definitely a problem. Please retest and let me know if it's better.

bluairis (approved):
Everything works! Looks great!

PR #677, opened 2023-03-23
Move page button is invisible

### LINKED ISSUE #676: Move page button is invisible
## Expected behavior
When trying to move a page in Wagtail the page editor should be able to select a new parent page for the page being moved. This is done by clicking on a button in the move page dialog.

## Actual behavior
The button for moving pages is invisible!

image
 _Note: There should be a set of buttons next to where it says, "New parent page"._

The page should look like this:

image

## Steps to reproduce
- Using a desktop browser click on the three dots next to any standard page  or internet plain page in the admin and select "Move".

You will see the move page dialog but there will be no way to move the page. If you hover around to the right of the dialog you will find an invisible button.

## Related code
This is a problem with core Wagtail. It's due to a misapplied CSS4 media query that basically says, "if a device supports hover, set the opacity to 0". Because of this, the problem only affects desktop computers and other devices with a hover state. The problem does not manifest on mobile devices.

```
@media (hover: hover)
.chooser__actions {
  opacity: 0;
}
```

## Other information:**
This problem effects desktop computers. It does not effect mobile devices. It can likely be fixed with the `insert_global_admin_css` hook: https://docs.wagtail.org/en/stable/reference/hooks.html#insert-global-admin-css

bbusenius:
Fixes #676 

**Changes in this request**
- Added admin css to override the bad rule.

`./manage.py compress` and restart dev server to test.

### DIFF
--- a/static/css/admin.css
+++ b/static/css/admin.css
@@ -1,1 +1,5 @@
 /* Add styles for overriding things in the admin. */
+
+.chooser__actions {
+    opacity: 100;
+}


bufordrat (approved):
Beautifully done!  It's pretty cool that we're able to patch a problem in the actual Wagtail project so simply.

PR #693, opened 2023-06-16
Page not found should return 404

### LINKED ISSUE #692: Page not found should return 404
Page not found view is returning a `200` for pages that aren't found. This should return a `404`.

bbusenius:
Fixes #692

**Changes in this request**
- Make the page not found throw a 404 instead of a 200.


bufordrat (approved):
Boo yah, totally psyched we're putting this change in.  404 status code shows up in my browser when I go to the 404 page.

PR #699, opened 2023-07-21
new mail aliases pr to fix the merge and requested changes

### LINKED ISSUE #658: Incorporate Mail Aliases Page into Wagtail Site
- Currently located at [http://www3.lib.uchicago.edu/cgi-bin/mail-aliases/](url)
- We want to build a page in wagtail with the same information from a json export located in the motacilla file system

bluairis:
Fixes #658 

**Changes in this request for Mail Aliases Page**

- Created a new branch based off of master to deal with the accidental merge
- Also fixes the issues that were listed in the previous PR

### DIFF
--- a/intranethome/views.py
+++ b/intranethome/views.py
@@ -10,6 +10,7 @@ from django.http import Http404
 from django.shortcuts import render
 from library_website.settings import MAIL_ALIASES_PATH
 from wagtail.models import Site
+import logging
 
 
 def reading_and_converting(name_of_file):
@@ -25,19 +26,22 @@ def reading_and_converting(name_of_file):
         a dictionary in the format {'error':'error'} to indiate that the file was not found
     """
 
-    #try:
-    f = open(name_of_file, "r")
-   # except FileNotFoundError:
-        # todo: log this error
-    #    return {"error": "error"}
-    #except IOError:
-        # todo: log this error
-     #   return {"error": "error"}
-    #else:
-    json_text = f.read()
-    f.close()    
-    data_from_file = json.loads(json_text)
-    return data_from_file
+    try:
+        f = open(name_of_file, "r")
+    except FileNotFoundError as e:
+        logging.log(str(e))
+        return {"error": "error"}
+    except IOError as e:
+        logging.log(str(e))
+        return {"error": "error"}
+    except JSONDecodeError as e:
+        logging.log(str(e))
+        return {"error": "error"}
+    else:
+        json_text = f.read()
+        f.close()    
+        data_from_file = json.loads(json_text)
+        return data_from_file
 
 
 def helper_function_order_number_aliases_last(all_aliases):
--- a/intranethome/views.py
+++ b/intranethome/views.py
@@ -28,6 +28,10 @@ def reading_and_converting(name_of_file):
 
     try:
         f = open(name_of_file, "r")
+        json_text = f.read()
+        f.close()    
+        data_from_file = json.loads(json_text)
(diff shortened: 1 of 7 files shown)


library_website/urls.py
```diff
@@ -53,6 +54,7 @@
     ),
     url(r'^about/news-events/events/$', events_view, name='events'),
     url(r'^collex/$', collection_view, name='collection'),
+    url(r'^mailaliases/', mail_aliases_view, name='mail_aliases'),
```
bbusenius, line 57: I'm still able to access the page at http://wwwdev:8000/mailaliases/. I might be misremembering but I recall that wasn't possible in your original version.

intranethome/views.py
```diff
@@ -0,0 +1,342 @@
...
+        # option 1: single key dictionary with key = 'email' and value with a local or gobal email
+        #   {'email': 'postmaster'}
+
+        # option 2: list of single dictionaries with  key = 'email' and value with a single local or global email, in one of 3 formats
+        #   [{'email': 'lhauglan@midway.uchicago.edu'}, ..., {'email': 'd-larsen@uchicago.edu'}]
+        #   [{'email': 'fierrom@uchicago.edu (Maria Fierro)'}, ..., {'email': 'hartj@uchicago.edu (Jenny Hart)'}]
+        #   [{'email': 'Ariel Erbacher\t\t<aaescotese@uchicago.edu>'}, ..., {'email': 'Amy Boucher\t\t<aebouch@emory.edu>'}]
+
+        # option 3: dictionary with key = 'note'
+        #   {'note': 'emails go to command: "|/usr/local/mailman/mail/mailman post access-ip'}
```
bbusenius, line 137: This is more clear for sure. I don't think the # symbols are adding anything though. I would remove those. It might be more clear if you added something like, "Possible input looks like this:" (or something to that effect, after "converted into the correct format". I really like how it says "(option 1 or 3)" and (option 2)" in the `notes_or_emails` argument description though, don't change that. That is very helpful!

bbusenius (changes requested):
There are still a few unresolved issues:

- Delete the default comment in tests.
- The except statement still doesn't specify what kind of exception.
- [New comment] further slight enhancement to docstring (but looking much better now).

library_website/urls.py
bluairis: When I test this, I am not able to access my page from either wwwdev:8000/mailaliases not from liblet.lib.uchicago.edu/mailaliases

intranethome/views.py
```diff
@@ -0,0 +1,333 @@
...
+
+    else:
+
+        # grabs /mailaliases/*the_filter_value*
+        url = request.get_full_path()
+
+        # pulls *the_filter_value* out
+        try:
+            alias_filter = re.search("\/mailaliases\/([^\/]*)\/?", url)[1]
+        except:
```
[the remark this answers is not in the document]
bluairis: I wasn't exactly sure how to handle this situation. Basically, the except case is triggered when there is nothing after the slash (loop.lib.uchicago.edu/mailaliases/). When I set the filter to "", it means that there is no filter and all of the data gets displayed. Therefore, there is only 1 case that would qualify for the except which is nothing being after the slash. How would you prefer that I approach this?

bbusenius (changes requested):
Pulling back on the approval until you sort out logging and tests.

bbusenius (approved):

bufordrat (approved):
Looking good.

PR #701, opened 2023-07-25
fixed the footer and right side header links on loop homepage

### LINKED ISSUE #700: Updating Mail Alias Site Links
The links on loop need to be changed to point to the new mail aliases site

bluairis:
Fixes #700 

**Changes in this request**

- Fixes the link at the footer of loop
- Fixes the link on the loop homepage on the right hand side


bbusenius (changes requested):
Can we make these site relative links, e.g. `/mailaliases/` so they work in dev too? The old links were absolute because they were going to something offsite (Apache exception).

bbusenius (approved):
Looks good! Thank you!

bufordrat (approved):
w00t

PR #704, opened 2023-07-28
Upgrade Django and Wagtail

### LINKED ISSUE #703: Upgrade Django and Wagtail
Upgrade Django to `4.1.10` and Wagtail to `4.1.6` LTS release.

bbusenius:
Fixes #703 

**Changes in this request**
- Upgrades Django to `4.1.10`.
- Upgrades Wagtail to `4.1.6` (LTS version)
- Adds documentation to `Vagrantfile` to help with onboarding.


Vagrantfile
```diff
@@ -53,6 +53,51 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
     up_message = <<-MSG
 
+    Linting for Python and React
+    ============================
+
+    Lint Python files with isort, autopep8, and black. Lint React Javascript
+    using eslint. This can be done here using Vim or in your editor of choice.
+    Look at ~/.vimrc to see relevant settings (you will need to configure your
+    editor in a similar way).
+
```
bufordrat, line 63: Might be helpful to include linting commands here.  So:

```
$ isort path/to/source_file.py
$ autopep8 --in-place path/to/source_file.py
$ black path/to/source_file.py
```

base/migrations/0016_alter_intranetindexpage_editor_and_more.py
```diff
@@ -0,0 +1,145 @@
...
+                related_name="%(app_label)s_%(class)s_maintainer",
+                to="staff.staffpage",
+            ),
+        ),
+        migrations.AlterField(
+            model_name="localmedia",
+            name="caption_language",
+            field=models.CharField(
+                choices=[
+                    ("af", "Afrikaans"),
```
bufordrat, line 60: Huh, interesting to see our Wagtail site already has a dictionary mapping ISO language codes to language names.  Did not know that.

bufordrat (approved):
Great!  Fresh new dev environment in Vagrant works perfectly on my machine.  My only suggestion is to add those example linting commands to the help message for linting, but I'll leave it to your discretion about whether to put those in.

base/migrations/0016_alter_intranetindexpage_editor_and_more.py
bbusenius: Yes, there is some of that. I think there's a Django module you can just import and use as well. I'm fuzzy on the details but I recall something like that.

Vagrantfile
bbusenius: Good idea. I never run them like that but it makes sense that somebody would. I'll add them.

PR #709, opened 2023-09-26
 Fix dev environment issues

### LINKED ISSUE #708: Fix dev environment issues
## Issues
- Elaticsearch in dev doesn't match what we use in production and in `requirements.txt`. This breaks the installation.
- NodeJS installation now throws a warning during the build process.

The Elasticsearch installed in the dev environment needs to be updated to `7.x` to match what is expected in `the requirements.txt` and `NodeJS` needs to be installed differently now.

bbusenius:
Fixes #708

**Changes in this request**
- Update the dev environment to Elasticsearch 7.
- Update to Elasticsearch 7 in GitHub Workflows.
- Change the way we install NodeJS to eliminate new warning.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,7 +33,7 @@ jobs:
         sudo sysctl -w vm.max_map_count=262144
     - uses: elastic/elastic-github-actions/elasticsearch@master
       with:
-        stack-version: 6.8.13
+        stack-version: 7.17.13
 
     - uses: actions/checkout@v3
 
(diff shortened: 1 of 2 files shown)


bbusenius:
Looks like tests are failing here because I need to update Elasticsearch in CI as well. I'm looking into it.

bufordrat (approved):
Excellent!  Had to struggle a bit with VirtualBox kernel modules, as you know, but it looks like everything is working.

PR #727, opened 2024-03-01
Links to events from calendar return 404s

### LINKED ISSUE #726: Links to events from calendar return 404s
## Problem
Event links on the [Library calendar](https://www.lib.uchicago.edu/about/news-events/events/) return 404s. It appears that LiveWhale used to return absolute links but now returns site relative links. This is causing that once went to events.uchicago.edu to go to www.lib.uchicago.edu (where the pages don't exist).

bbusenius:
Fixes #726

**Changes in this request**
- Added config option for an events feed base url.
- Updated the config and calendar code to use the new config.
- Linted.

NOTE: One of the edited files had never been linted. [Look at the first commit to see the actual changes](https://github.com/uchicago-library/library_website/commit/e7def5a70998688199ea58dd49699b10334747d5). The rest is just linting.

### DIFF
--- a/events/utils.py
+++ b/events/utils.py
@@ -1,11 +1,12 @@
+import itertools
+import re
 from datetime import datetime, timedelta
 from http.client import HTTPConnection, HTTPSConnection
 from urllib.parse import parse_qs, urlparse
 from xml.etree import ElementTree
+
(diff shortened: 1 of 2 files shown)


bufordrat (approved):
Very nice!

PR #746, opened 2024-04-05
728 mail aliases needs a new test suite

### LINKED ISSUE #728: Mail aliases needs new test suite
Following #724, the old test suite for the `intranethome` app on the library website is now out of date.  For this issue, please write an updated test suite for the mail aliases page with similar coverage to the old mail aliases page.

bluairis:
Fixes #728 

**Changes in this request**

- I added a few tests for the mail aliases page (I did not test each_list because it needs to be fixed pretty soon)
- I made one change to the filter function because the test made me realize that it was removing aliases like "MAILER-DAEMON" from "m" filter value because it was only looking for lowercase first values
- Tests are passing.

### DIFF
--- a/intranethome/tests.py
+++ b/intranethome/tests.py
@@ -7,6 +7,7 @@ from intranethome.views import (
     get_sorted_aliases,
     formatting,
     filter_by_value,
+    parse_file,
 )
 
 class test_mail_aliases_view(SimpleTestCase):
@@ -83,4 +84,5 @@ class test_mail_aliases_view(SimpleTestCase):
         self.assertEqual(filter_by_value(aliaslist, "number"), ["1Candy", "2Candy", "1dumplings"])
 
     def test_missing_MAIL_ALIASES_PATH(self):
-        self.assertEqual(True, True)
+        path = "bad_path.json"
+        self.assertEqual(parse_file(path), -1)
--- a/intranethome/tests.py
+++ b/intranethome/tests.py
@@ -1,3 +1,50 @@
 from django.test import SimpleTestCase
 
 # TODO: write new tests to reflect the updated code in views.py
+
+from intranethome.views import (
+    formatting,
+    comparison,
+)
+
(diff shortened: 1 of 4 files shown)


intranethome/tests.py
```diff
@@ -1,3 +1,83 @@
...
+        alias2 = "BAD_APPLE"
+        alias3 = "1Candy"
+        alias4 = "2Candy"
+        alias5 = "1dumplings"
+        aliaslist.append(alias1)
+        aliaslist.append(alias2)
+        aliaslist.append(alias3)
+        aliaslist.append(alias4)
+        aliaslist.append(alias5)
+        self.assertEqual(filter_by_value(aliaslist, ""), aliaslist)
```
bbusenius, line 81: It's good to see you testing for the empty string case.

bbusenius (changes requested):
These tests are a good start. I have a couple of suggestions:

1. It would be nice if the names of the test functions conveyed more information about what you're testing about the functions. The Django convention is to use long descriptive function names in tests, rather than docstrings, however, there's nothing wrong with adding a docstring if this isn't possible or if you want to be more descriptive. [Part 5 of the Django tutorial](https://docs.djangoproject.com/en/5.0/intro/tutorial05/) has good information about this.
2. It would be really good to have a test that covers the case for when the MAIL_ALIASES_PATH file isn't found.

bufordrat (changes requested):
Looks great, tests pass on my machine.  I don't have any suggestions other than that I agree with the suggestion to have the site for the case where there is no file at `MAIL_ALIASES_PATH`, and then to also write a test confirming that the site doesn't blow up with an internal server error when that file is missing.  With-finalizing in `parse_file` looks good.

bufordrat:
These changes were incorporated into the code by way of PR #766.  Thank you @bluairis for all your work on this.

bufordrat:
Closing the issue...

PR #761, opened 2024-04-26
753 inconsistet fonts and spacing on loop

### LINKED ISSUE #753: Inconsistet fonts and spacing on Loop
## Problem
On [some Loop pages](https://loop.lib.uchicago.edu/groups/technical-services-policy-group/technical-services-policies/policy-e-holdings-and-urls/), both the font size and spacing is irregular. The line spacing and font size is different between paragraphs and lists. Also, paragraphs have a constrained width while lists do not:

image

## Other information:
https://www.w3.org/WAI/tutorials/page-structure/styling/#line-length

### LINKED ISSUE #754: sidebar and footer content are mixed on mobile
## Problem description
David Bottorff: "on an iOS device in safari on at least some pages like https://www.lib.uchicago.edu/scrc/exhibits/, the left hand nav overlaps with the footer and can’t be clicked upon properly"
*See screenshot at the bottom.

## Possible solutions
One possible solution is to include the footer in the `row-offcanvas` div where the remainder of the page content lies.
Another solution would be to make the sidebar completely hide the footer. The text in the footer somehow hovers above the sidebar, so this rule would fix this. We can also consider having the button that toggles the sidebar sticky, so that one doesn't need to scroll long pages to bring the sidebar in and out.
```
div#sidebar.sidebar-offcanvas {
    z-index: 1;
}
```

## Anex
image

vitorgomateus:
Fixes #753.
Fixes #754.

**Changes in this request**
- create consistency in the content block width by changing the limit from the `p` rule to the general `.block-paragraph`.
- add `z-index` to the left sidebar so that its content doesn't get visually mixed with the footer.
- new logo's maroon has low contrast with loop's green so it was changed to the all-white logo.

### DIFF
--- a/base/static/base/css/loop.css
+++ b/base/static/base/css/loop.css
@@ -129,8 +129,7 @@ h1 + h2, h2 + h3, h3 + h4, h4 + h5, h5 + h6 {
  */
 p {
   font-size: 1.1em;
-  line-height: 1.7em;
-  max-width: 46em;
+  line-height: 1.7em;  
   margin-bottom: 0.5em; }
 
 .template-news-page .rich-text {
@@ -268,6 +267,20 @@ figure.center-block {
  *  Content
  * --------------------------------------------------
  */
+
+.block-paragraph {
+  max-width: 46em;
+}
+.block-paragraph p,
+.block-paragraph a,
+.block-paragraph li {
+  font-size: 16px;
+  line-height: 22px;
+}
+.block-paragraph a{
+  overflow-wrap: break-word;
+}
+
 .sw-toc {
   display: flex;
   flex-wrap: wrap;
--- a/base/static/base/css/loop.css
+++ b/base/static/base/css/loop.css
@@ -267,17 +267,16 @@ figure.center-block {
  *  Content
  * --------------------------------------------------
  */
-
-.block-paragraph {
+    
+.block-paragraph,
+.media-body{
   max-width: 46em;
 }
-.block-paragraph p,
-.block-paragraph a,
-.block-paragraph li {
-  font-size: 16px;
+[data-block-key]{
+  font-size: 15px;
   line-height: 22px;
 }
-.block-paragraph a{
+[data-block-key] a{
   overflow-wrap: break-word;
 }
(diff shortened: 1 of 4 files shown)


bbusenius (changes requested):
The logo looks great and I like your use of the consolidated web-resources folder. I'm still seeing font problems on loop though. The lists still aren't constrained and the font size is smaller:

image

I emptied the cache and hard-refreshed so I don't think I'm seeing a caching issue.

vitorgomateus:
I didn't notice that the Loop SCSS needs to be compressed separately. I wasn't able to run Gulp, so I made 'compressed' changes manually. I also fixed the difference in font-size and line-height, and added a `overflow-wrap: break-word;` because I saw a long link in this particular example page.

bbusenius:
This is looking a lot better. I see the text is constrained now and the line height looks better. I still see a couple of issues though.

1. The list font is still smaller than the paragraph text. I think this is because the paragraph has `font-size: 1.1em` but the list is just inheriting `font-size: 14px` from the body.
2. I think we need to get rid of this `gulp` dependency and move this SASS into the compress framework.

Maybe if we address the first issue in this PR, we can open a new PR for the second. Otherwise, we could address them both here. I don't have a strong preference.

vitorgomateus:
It seems these new rules do not apply on the news page type. I'll make new rules to include those as well.
I will create a new issue to get rid of the gulp dependency, or at least create consistency between lib and loop because gulp would be helpful for dev, but it has to work with compress. So I wanna take a little extra time on it.

vitorgomateus:
I've tested these pages and it seems the new styles will be consistent:
https://looplet.lib.uchicago.edu/groups/technical-services-policy-group/technical-services-policies/policy-e-holdings-and-urls/
https://looplet.lib.uchicago.edu/departments/administration-and-operations/assessment/
https://looplet.lib.uchicago.edu/documentation/website-authoring/sidebar-promote/library-news/

bbusenius:
Nice, this looks good to go!

bbusenius (approved):

PR #773, opened 2024-06-26
Make it possible to spin up the dev site without Elasticsearch or NodeJS

### LINKED ISSUE #769: Make it possible to spin up the dev site without Elasticsearch or NodeJS
## Issue
Use environment variables to make it possible to spin up a dev version of the site that doesn't use Elasticsearch, Java, or NodeJS. Since we don't need these features for the vast majority of development and they slow down the build process, they should be optional. Is there anything else we should make optional?

## Other information:
- https://stackoverflow.com/questions/14124234/how-to-pass-parameter-on-vagrant-up-and-have-it-in-the-scope-of-vagrantfile
- https://stackoverflow.com/questions/26811089/vagrant-how-to-have-host-platform-specific-provisioning-steps

bbusenius:
Fixes #769

**Changes in this request**
- Ability to build the site without Elasticsearch.
- Ability to build the site without NodeJS
- Removed `Poetry` and `Fabric` dependencies in the build (these were never used).

To test run `ELASTICSEARCH=false NODEJS=false vagrant up` and put 
```
WAGTAILSEARCH_BACKENDS = {
    'default': {
        'BACKEND': 'wagtail.search.backends.database',
    }
}
```
in your `local.py`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,20 @@
 3. The following commands run automatically after `vagrant ssh` to activate the virtualenv and move to the working directory: `source lw/bin/activate && cd /vagrant/`
 4. Start the Django dev server: `./manage.py runserver 0.0.0.0:8000`
 
+### Build the site withouth Elasticsearch or NodeJS
+Since the majority of development tasks don't require Elasticsearch or NodeJS it is often faster and more desireable to build the site without these things. This can be accomplished by running `vagrant up` with the `ELASTICSEARCH` or `NODEJS` environment variables set to `false`:
+```
+ELASTICSEARCH=false NODEJS=false vagrant up
+```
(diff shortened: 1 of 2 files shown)


bufordrat (approved):
Worked perfectly on my machine, can you believe it?  ;)

At first I was thrown off by the fact that the Python `elasticsearch` library got installed, but when I took a closer look, I saw that the conditional logic only applies to Elasticsearch at the OS level, which makes sense.  When I tested the site out it was just using Wagtail search.  

Thanks again; this no-elasticsearch feature seems like it'll be really handy for testing in dev mode.

bbusenius:
Yes, the Python package is still installed, just not the beastly ES, also no Java which is probably the real feature.

PR #781, opened 2024-08-01
don't repetitively load jquery on staff page

### LINKED ISSUE #780: don't repetitively load jquery on staff page
@vitorgomateus recently noticed that our staff page was loading jQuery 3.3.1 every time it displayed a "schedule an appointment" button.  Oops.

For this issue, please move the jQuery load to the top of the staff page listing.

To see the repetitive jQuery load, navigate here and view source:
https://www.lib.uchicago.edu/about/directory/

bufordrat:
Fixes #780.

Just moves the jQuery include from the `staff` app to the `unit` app so that jQuery 3.3.1 loads once.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ library_website/settings/database.db
 /.vagrant/
 
 src
+
+**/*~
--- a/staff/templates/staff/libcal_button.html
+++ b/staff/templates/staff/libcal_button.html
@@ -1,7 +1,6 @@
 {% load staff_tags %}
 
 {% if libcal_id %}
-<script src="//ajax.googleapis.com/ajax/libs/jquery/3.3.1/jquery.min.js"></script>
     <script>
 	jQuery.getScript("https://api3.libcal.com/js/myscheduler.min.js", function() {
 	    var id_num = "{{ libcal_id }}";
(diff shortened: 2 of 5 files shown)


bbusenius (approved):
This looks good to me. I'm surprised we don't have a JS block in the base page template but I guess it's never come up before. I see we conditionally load JQuery on the hours page there (different version, I'm sure). Anyhow, I don't think either of those approaches are necessary in this case. Tests are passing. I'm happy.

vitorgomateus (approved):
Looks good to me.

vitorgomateus:
I was getting ready to merge this, but it seems the "Schedule an Appointment" button is loaded in multiple locations, and I wonder if the jquery script is loaded in all of them? I'm having a hard time testing it locally with dev data.
Do we need to make sure to load the jquery in:
`library_website\base\templates\base\public_base.html`
`library_website\public\templates\public\staff_public_page.html`
`library_website\staff\templates\staff\staff_directory_listing.html`

vitorgomateus:
Well, after looking to see if the button was loaded in more places, I ended up pushing some updates. Sorry for kind of taking over @bufordrat . I also added the updates from master so we need another review @bbusenius

bbusenius (approved):
Good catch! This looks good to me. Thanks for dealing with the merge conflicts and getting this thing out the door!

PR #803, opened 2025-01-23
Inverted Groups

### LINKED ISSUE #802: Inverted Groups
Groups and subgroups are inverted on the [Loop groups page](https://loop.lib.uchicago.edu/groups/). This bug was introduced with [PR 799](https://github.com/uchicago-library/library_website/pull/799). If you look at the url on the groups and subgroups, you can see that the parents are actually children and the children are actually parents.

Image

bbusenius:
Fixes #802 

**Changes in this request**
- Fixed bug that was causing Loop groups and subgroups to be inverted.

### DIFF
--- a/group/models.py
+++ b/group/models.py
@@ -613,12 +613,10 @@ class GroupIndexPage(BasePage):
         # new levels or using the existing ones if they're already there
         # from previous group pages.
         for grouppage in (
-            GroupPage.objects.live().child_of(self).filter(is_active=True)
+            GroupPage.objects.live().descendant_of(self).filter(is_active=True)
         ):
             ancestors = (
-                [self]
-                + list(GroupPage.objects.descendant_of(grouppage))
-                + [grouppage]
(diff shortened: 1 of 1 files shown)


vitorgomateus:
```
for index_child in index_children:
            for inactive_child in inactive_children:
                if inactive_child.is_descendant_of(index_child):
                    inactive_children = inactive_children.not_descendant_of(index_child)
```
Is this to avoid circular dependency?
It looks good to me, I don't understand the code completely.

bbusenius:
@vitorgomateus, yes, it's very confusing code (all the groups code is confusing). That part is necessary to filter out the group pages that are children of other `GroupIndexPages` since other `GroupIndexPages` can reside under themselves. Without that, it would display all `GroupPages` in the entire tree. But we only want those that aren't listed under any other `GroupInmdexPage`. It's not elegant.

vitorgomateus (approved):
looks good to me

bufordrat (approved):
Looks great!  This type of tree-pruning code is super fiddly/hairy, but it works well on my machine in dev.

Issue #770, opened 2024-06-14
Re-enable ID ResolveTests


bbusenius:
## Issue

The ID [Resolve tests are commented out](https://github.com/uchicago-library/library_website/blob/master/public/tests.py). I think we want to re-enable these, if not, we should delete them.

bufordrat:
Sorry, just noticed this issue.  I think it's safe to bring them back.  Will take a look.

PR #806, opened 2025-02-11
bring the `idresolve` tests back

bufordrat:
Bring the `idresolve` unit tests back, as per #770.


bbusenius (approved):
Well, tests are passing, so I'd say we're good! Thanks for looking into this!

PR #818, opened 2025-03-25
817 fix finding aids crash when logged in

### LINKED ISSUE #817: fix finding aids crash when logged in
Currently, the library website crashes in the browser, possibly due to some kind of memory leak, when the user clicks on the Finding Aids page and then logs in via Okta:

https://www.lib.uchicago.edu/scrc/finding-aids/

The exact cause of this behavior is unknown.  However, it appears to be triggered by the combination of logging in and having to display a large-ish number of DOM elements (in this case, just over 1000 links to our Finding Aids website).  

For this issue, adjust the UI with pagination so that there is an upper limit on the number of DOM elements the Wagtail site has to display while a user is logged in.

bufordrat:
Fixes #817.

This PR avoids the in-browser memory leak by:

- adding an "all" button to the alphabetic browse links, which takes the user to a paginated listing of all finding aids pages
- retaining non-paginated listing of all title alphabetic browses
- retaining non-paginated listings of topic and digital-content-related browses
- retaining the old behavior in the face of searches

In addition, it simplifies the conditional logic in the `findingaids/templates/findingaids/finding_aids_page.html` template into something more linear, hopefully making the template easier to maintain in the long run.

### DIFF
--- a/findingaids/models.py
+++ b/findingaids/models.py
@@ -220,11 +220,7 @@ class FindingAidsPage(PublicBasePage):
         topiclist = []
         thistopiclist = []
 
-        if view == 'all':
-            pass
-        elif view == 'title':
-            pass
-        elif view == 'digitized':
+        if view == 'digitized':
             digitizedlist = get_digitized_content()
         elif view == 'topics':
             topiclist = get_topic_list(all_topics)
--- a/findingaids/models.py
+++ b/findingaids/models.py
@@ -26,23 +26,23 @@ class FindingAidsPage(PublicBasePage):
         def get_marklogic_xml(route):
             try:
                 r = urllib.request.urlopen(marklogic_url(route))
-                xml_string = r.read().decode("utf-8")
+                xml_string = r.read().decode('utf-8')
                 return xml_string
             except urllib.error.URLError:
-                return ""
+                return ''
 
         def get_browses():
             browses = []
 
             route = (
-                "/admin/gimme.xqy" "?collection=institution%2FUniversity%20of%20Chicago"
+                '/admin/gimme.xqy" "?collection=institution%2FUniversity%20of%20Chicago'
             )
 
             xml_string = get_marklogic_xml(route)
 
             try:
                 e = ElementTree.fromstring(xml_string)
-                for div in e.find("body").findall("div"):
+                for div in e.find('body').findall('div'):
                     span = div.findall("span")
                     if span[0].text and span[1].text:
                         browses.append([span[0].text, span[1].text])
@@ -66,19 +66,19 @@ class FindingAidsPage(PublicBasePage):
 
             digitized = []
 
-            route = "/admin/gimmeDigitalEADIDs.xqy"
+            route = '/admin/gimmeDigitalEADIDs.xqy'
 
             xml_string = get_marklogic_xml(route)
 
             try:
                 e = ElementTree.fromstring(xml_string)
-                for div in e.find("body").findall("div"):
-                    span = div.findall("span")
+                for div in e.find('body').findall('div'):
+                    span = div.findall('span')
                     digitized.append(
                         [
-                            span[0].find("eadid").text,
+                            span[0].find('eadid').text,
                             span[1].text,
-                            span[2].find("abstract").text,
+                            span[2].find('abstract').text,
                         ]
                     )
             except ElementTree.ParseError:
@@ -101,14 +101,14 @@ class FindingAidsPage(PublicBasePage):
             if exactphrase:
                 route = "/request.xqy?" + urllib.parse.urlencode(
                     {
-                        "action": "search",
-                        "collection": "project/SCRC",
-                        "q": '"' + searchq + '"',
+                        'action': 'search',
+                        'collection': 'project/SCRC',
+                        'q': '"' + searchq + '"',
                     }
                 )
             else:
                 route = "/request.xqy?" + urllib.parse.urlencode(
-                    {"action": "search", "collection": "project/SCRC", "q": searchq}
+                    {'action': 'search', 'collection': 'project/SCRC', 'q': searchq}
                 )
 
             xml_string = get_marklogic_xml(route)
@@ -120,9 +120,9 @@ class FindingAidsPage(PublicBasePage):
                 ):
                     searchresults.append(
                         {
-                            "eadid": div.find("div[@class='eadid']").text,
-                            "title": div.find("div[@class='project']").text,
-                            "abstract": div.find("div[@class='abstract']").text,
+                            'eadid': div.find("div[@class='eadid']").text,
+                            'title': div.find("div[@class='project']").text,
+                            'abstract': div.find("div[@class='abstract']").text,
                         }
                     )
             except (AttributeError, ElementTree.ParseError):
@@ -139,8 +139,8 @@ class FindingAidsPage(PublicBasePage):
 
             try:
                 e = ElementTree.fromstring(xml_string)
-                for div in e.findall("div"):
-                    subject = div.find("subject").text
+                for div in e.findall('div'):
+                    subject = div.find('subject').text
 
                     if subject is None:
                         continue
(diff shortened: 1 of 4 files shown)


findingaids/models.py
```diff
@@ -207,32 +198,56 @@ def get_topic_list(topics):
...
-        topiclist = []
-        if view == 'topics':
-            topiclist = get_topic_list(all_topics)
+        cached_browses = cache.get("finding_aids_all_browses")
+
+        if cached_browses:
+            all_browses = json.loads(cached_browses)
+        else:
+            all_browses = get_browses()
+            cache.set("finding_aids_all_browses", json.dumps(all_browses))
```
bbusenius, line 207: I like this.

findingaids/templates/findingaids/finding_aids_page.html
```diff
@@ -3,189 +3,266 @@
...
+                <form action="." class="searchbox" method="get">
+                    <input type="search" required="" onkeyup="buttonUp();" class="searchbox-input" name="searchq" placeholder="Search finding aids"{% if searchq %} value="{{ searchq }}"{% endif %}/>
+                    <span class="searchbox-icon"><i title="search" class="fa fa-search"></i>
+                        <input type="submit" class="searchbox-submit" style="background-color: transparent; color: transparent;"/>
+                    </span>
+                </form>
+            </div>
+            <div class="col-md-12">
+                <!-- just padding -->
+                <p/>
```
bbusenius, line 27: I think this is technically syntactically valid but probably not good for accessibility. I think it will be interpreted as an open p: `<p>`, which is technically fine but not ideal. Either a SASS fix or `<br/>` would be better.

findingaids/templates/findingaids/finding_aids_page.html
```diff
@@ -3,189 +3,266 @@
...
-                </th>
-              </tr>
-            </thead>
-          </table>
+            <!-- case 1: user has just entered a search term -->
+            {% if searchq %}
+                {% if searchresults %}
+                    <div class="col-md-12">
+                        <!-- just padding -->
+                        <p/>
```
bbusenius, line 35: Same as prior comment.

findingaids/templates/findingaids/finding_aids_page.html
```diff
@@ -3,189 +3,266 @@
...
+                            <li><a href="{{ root_link }}?view=title&browse=all&page={{ page_obj.next_page_number }}">&raquo;</a></li>
+                        {% else %}
+                            <li class="disabled"><span>&raquo;</span></li>
+                        {% endif %}
+                    </ul>
+                    <div class="text-primary col-md-12">
+                        page {{ page_obj.number }} of {{ num_pages }}
+                        <div class="col-md-12">
+                            <!-- more padding -->
+                            <p/>
```
bbusenius, line 136: Same as prior comment.

findingaids/templates/findingaids/finding_aids_page.html
```diff
@@ -3,189 +3,266 @@
...
+                                <li>
+                                    <a href="{{ root_url }}?view=title&amp;browse={{ b|urlencode }}">{{ b }}</a>
+                                </li>
+                            {% endif %}
+                        {% endfor %}
+                    </ul>
+                </center>
+                <div class="col-md-12">
+                    <!-- just padding -->
+                    <p/>
```
bbusenius, line 169: Same as prior comment.

findingaids/templates/findingaids/finding_aids_page.html
```diff
@@ -3,189 +3,266 @@
...
+                        <tbody>
+                            {% for b in browses %}
+                                <tr>
+                                    <td>
+                                        <strong>
+                                            <a target="_blank" href="http://www.lib.uchicago.edu/e/scrc/findingaids/view.php?eadid={{ b.0|urlencode }}">{{ b.1 }}</a>
+                                        </strong>
+                                    </td>
+                                </tr>
+                            {% endfor %}
```
bbusenius, line 190: This could be written as `{% for b0, b1 in browses %}` as well. Also, I'm wondering if we could get the findingaids base url from site settings or ini. I won't block for that though. It's not directly related to the issue you're solving. I just wanted to point it out. But I see this url is repeating. If we're not going to make it configurable, maybe we should at least put it in a constant.

findingaids/templates/findingaids/finding_aids_page.html
bufordrat: Whoops; thanks for catching that; I intended to replace that with something more respectable and then forgot.

findingaids/templates/findingaids/finding_aids_page.html
bufordrat: Great suggestion; thanks.

findingaids/templates/findingaids/finding_aids_page.html
```diff
@@ -3,189 +3,263 @@
...
+                <form action="." class="searchbox" method="get">
+                    <input type="search" required="" onkeyup="buttonUp();" class="searchbox-input" name="searchq" placeholder="Search finding aids"{% if searchq %} value="{{ searchq }}"{% endif %}/>
+                    <span class="searchbox-icon"><i title="search" class="fa fa-search"></i>
+                        <input type="submit" class="searchbox-submit" style="background-color: transparent; color: transparent;"/>
+                    </span>
+                </form>
+            </div>
+	    <div class="col-md-12" style="padding-top:1em!important">
+	        <!-- just padding -->
+	    </div>
```
vitorgomateus, line 27: I would reorganize to something like this, and remove the closing `</article>` tag. The search does not happen within each browse but separate from each, so to me having the search box above the tabs makes that more obvious. I would also remove the alphabets while not browsing By Title because they effectively change to the By Title tab, which I find confusing (you can go to the "K" of topics. or search topics), but that is not the concern here. BS3 does not use flexbox because this is a newer CSS feature, but it allows to more easily and consistently style this box, so I used it. This also seems to resolve boxing/div-ing issues that I found when testing. I didn't look much at the code within the conditionals but it seems to work well when testing.
```suggestion
<div class="row">
            <div class="col-xs-12 col-sm-6 col-sm-offset-3 distinct-search" style="display: flex ; flex-direction: column;">
                <form action="." class="searchbox" method="get">
                    <input type="search" required="" onkeyup="buttonUp();" class="searchbox-input" name="searchq" placeholder="Search finding aids"{% if searchq %} value="{{ searchq }}"{% endif %}/>
                    <span class="searchbox-icon"><i title="search" class="fa fa-search"></i>
                        <input type="submit" class="searchbox-submit" style="background-color: transparent; color: transparent;"/>
                    </span>
                </form>
                <div class="btn-group spaces-toggle" style="display: flex; flex-wrap: wrap;">
                    <a class="btn btn-list-toggle{% if view == 'title' %} active{% endif %}" href="?view=title&browse=all" style="flex-grow: 1;margin-bottom:.5em;">By Title</a>
                    <a class="btn btn-list-toggle{% if view == 'topics' %} active{% endif %}" href="?view=topics" style="flex-grow: 1;margin-bottom:.5em;">By Topic</a>
                    <a class="btn btn-list-toggle{% if view == 'digitized' %} active{% endif %}" href="?view=digitized" style="flex-grow: 1;margin-bottom:.5em;">With Digital Content</a>
                </div>
            </div>
        </div>
```

findingaids/templates/findingaids/finding_aids_page.html
```diff
@@ -3,189 +3,202 @@
...
+                    </tr>
+                </thead>
+            </table>
+        {% endif %}
 
-          <table class="table table-striped scrc-list">
+        <!-- case 2: user is browsing by title -->
+    {% elif page_obj and view == 'title' %}
+        <ul class="text-center pagination justify-content-center">
+            <li class="active">
```
vitorgomateus, line 66:
```suggestion
            {% if browse=='all' %}<li class="active">{% else %} <li> {% endif %}
```
Good to go!

vitorgomateus (approved):

bbusenius (approved):

PR #835, opened 2025-05-28
826 update click tracking to ga4

### LINKED ISSUE #826: Update click tracking to GA4
## Issue
Since the Universal Analytics to the Google Analytics 4 update the method to track link clicks has not been updated. This means that can't report on how users are clicking on different elements of the page.

The script `library_website\base\static\base\js\jquery.track-everything.js` needs to be rewritten to update to the new GA4 event pushing method.
```
gtag('event', event_name, {
	event_parameters
});
```
A list of all tracked links on the home page can be found [here](https://docs.google.com/spreadsheets/d/1L0ki9YyzXzWykCYyClvrwoqRyh6x9XlUda72py8gurk/edit?usp=sharing).

## New Script
Changing the GA4 structure would suffice, but we can also improve this code for performance, to track some extra information, and to track more pages.

### New Schema
```
event_name = "string"; // click, hover, dropdown, submit, tab-click
event_parameters = {
  "category": "string", // navigation links, quick links, tabs, left-sidebar, content
  "subcategory": "string", // widget-<name>, subject-guides, subject-specialists, or subject-collections
  "label": "string", // link name or text 
  "click_position": "int"
};
gtag('event', event_name, {
	event_parameters
});
```

### Requirements
- The script should get all the links on the page to track them into a GA4 event. 
- It will also need to check on each link for several details to determine which values are passed as event parameters. To determine these, there will be specific rules for specific contexts, for example if the link is within a navbar, a specific widget, a sidebar, the main content of the page, or on a specific page. Some links will have values asserted as data properties.
- **Event Delegation**: Instead of attaching event handlers to each link individually, consider using event delegation. Attach a single event listener to a common ancestor (e.g., the document body) and handle events as they bubble up. This can significantly improve performance, especially on pages with many links.
- **Debouncing/Throttling**: If you expect rapid successive events (e.g., hover events), consider debouncing or throttling to prevent excessive event firing.
- **Documentation**
- **Naming Conventions**
- **Edge Cases**
- **Conflict with screen readers**
- **Browser Compatibility**

vitorgomateus:
Fixes #826 

**Changes in this request**
- Completely rewrites the track everything script.
- It can now submit events to GA4.
- It can now track all clicks on VuFind. (vufind needs to include this script https://github.com/uchicago-library/vufind/tree/add-script-for-click-tracking)
- Code is much cleaner.
- Includes a debouncing function that was not working, but I left it there anyway.
- It still accepts specific html encoded event parameters, but it has a bunch of logic to track other cases.
- Improves homepage search widget tab tracking.

### DIFF
--- a/base/static/base/js/jquery.track-everything.js
+++ b/base/static/base/js/jquery.track-everything.js
@@ -132,22 +132,21 @@
                 params.event_subcategory = params.event_subcategory || 'list';
                 params.click_position = parseInt(link.getAttribute('data-ga-position'), 10);
             } else {
-                const parentLi = link.closest('li');
-                if (parentLi) {
+                const ancestor = link.closest('li');
+                // for any list
+                if (ancestor) { 
                     params.event_subcategory = params.event_subcategory || 'list';
-                    params.click_position = Array.from(parentLi.parentElement.children).indexOf(parentLi) + 1;
-                } else {
-                    const parentNews = link.closest('.news-wrap');
-                    if (parentNews) {
-                        params.event_subcategory = params.event_subcategory || 'news-list';
-                        params.click_position = Array.from(parentNews.children).indexOf(link.closest('.newsblock')) + 1;
-                    } else {
-                        const parentNews = link.closest('.news-stories');
-                        if (parentNews) {
-                            params.event_subcategory = params.event_subcategory || 'news-list';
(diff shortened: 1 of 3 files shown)


bbusenius (approved):
This looks good to me.

PR #843, opened 2025-07-02
do not delay behaviour on tab clicks

### LINKED ISSUE #842: Clicks on tabs should not scroll the page
PR #837 made tab clicks to scroll the page to the anchor, which should not happen.

vitorgomateus:
Fixes #842 

Clicks had their default behaviour delayed for sending custom events to GA4, but tab clicks do not need this so they should be excluded from this block.


bbusenius (approved):
Looks good!

PR #846, opened 2025-07-10
Fix event display issue with inconsistent GUIDs between feeds

### LINKED ISSUE #845: Fix event display issue with inconsistent GUIDs between feeds
## Problem

An event that was edited in LiveWhale is not appearing. This is because the URL was changed and the GUID was removed. TTRSS indexed the page with the GUID in the url and fails to match after the page was edited.

- TTRSS feed GUID format: `https://events.uchicago.edu/event/249475-reg-reads-book-club-summer-meeting`
- University feed GUID format: `https://events.uchicago.edu/event/reg-reads-book-club-summer-meeting`

The current implementation in `filter_university_entries_to_ttrss_entries` function only does exact string matching of GUIDs, which fails when the formats differ.

## Proposed Solution

Modify the `filter_university_entries_to_ttrss_entries` function in `events/utils.py` to use a more robust matching approach:

1. Extract event slugs from both feeds for comparison
2. Try exact matching first, then fall back to slug matching
3. Handle cases where event IDs are prepended with numbers

This will ensure that properly tagged events appear on all relevant kiosk pages regardless of GUID format inconsistencies.

## Testing

To test this fix:
1. Verify that events with multiple tags appear on all relevant kiosk pages
2. Verify that the "Reg Reads Book Club Summer Meeting" event now displays

bbusenius:
Title: Fix event display issue with inconsistent GUIDs between feeds

Fixes #845

Summary

Modified the event filtering logic to handle inconsistent GUID formats between University and TTRSS feeds.
- Added slug-based matching to handle events with different URL formats
- Fixed comparison between University feed's 'link' field and TTRSS feed's 'guid' field
- Implemented fallback matching for events with numeric IDs in URLs

Testing:
1. Verify that events with multiple tags (e.g., 'library' and 'science') now appear on all relevant kiosk pages
2. Check that the 'Reg Reads Book Club Summer Meeting' event appears on both the library homepage and crerar page

### DIFF
--- a/events/utils.py
+++ b/events/utils.py
@@ -227,13 +227,32 @@ def filter_university_entries_to_ttrss_entries(university_entries, ttrss_entries
         xml feed and the Tiny Tiny Rss feed.
     """
     ttrss_guids = []
+    ttrss_slugs = []
+
     for ttrss_entry in ttrss_entries:
-        ttrss_guids.append(ttrss_entry['guid'])
+        guid = ttrss_entry['guid']
+        ttrss_guids.append(guid)
+        # Extract the slug (part after the last slash)
+        slug = guid.split('/')[-1]
+        # If the slug contains a hyphen followed by text, extract just that part
+        if '-' in slug:
+            slug = slug.split('-', 1)[1]  # Split on first hyphen only
+        ttrss_slugs.append(slug)
 
     entries_out = []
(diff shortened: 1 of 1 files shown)


bufordrat (approved):
Works great; code looks good.

PR #855, opened 2025-08-29
851 correct custom labeling for click events

### LINKED ISSUE #851: Correct custom labeling for click events
*Problem*
Clicks in the search widget are not consistently tracked. As it is, it is not possible to filter all click, tab, and checkbox events happening on the search widget of the main page, and analyze these only or compare them with the navbar click, for example. While all navbar events have `event-category=main-nav-links`, the same is not true for the widget.

*Solution*
They should all have the Category = Search-Widget, the sub-category would be the tab, and the label the actual label. 
Maybe remove custom tracking labeling set in the HTML.
The script detects a click and if the element has custom labels for label, category, and subcategory, it will attribute those. Otherwise tries to establish from context, like the first parent with an ID, or by general section of the page like the Sidebar (depends on URL somewhat).
Adding more custom conditions to the script is a bad idea. Cleaning the HTML would be the best option.

*Bonus*
Adding a counter to the number of navbar dropdown events and submitting that count when the click event finally happens in the navbar (or now that I think of it, maybe anywhere else. It is a measure of how much the user paces around, which is useful). This could potentially be submitted as `click-position` as these are irrelevant in the navbar, but not always in other places.

vitorgomateus:
Fixes #851 .

**Changes in this request**
- Changes the click labeling significantly but makes it much easier to analyze it in GA. However data analysis will mostly not be backwards compatible.
- Moves most of click labeling into HTML.
- Better documentation.
- Adds a count of indecision on the navbar and search widget tabs to measure how many times the user changed dropdown or tab before selecting what they wanted.
- Adds options to search widget searches for better context.

### DIFF
--- a/base/templates/base/includes/header.html
+++ b/base/templates/base/includes/header.html
@@ -88,10 +88,10 @@
                         <li class="col-sm-6 ul-left"><span class="twocol-head" role="heading" aria-level="2" id="search_catalogs">Catalogs</span>
                             <ul class="list-unstyled" aria-labelledby="search_catalogs">
                                 <li role="separator" class="divider visible-xs" role="presentation"></li>
-                                <li><a href="/h/3" data-ga-label="Search &gt; Library Catalog">Library Catalog</a></li>
-                                <li><a href="/h/borrowdirect" data-ga-label="Search &gt; BorrowDirect">BorrowDirect</a></li>
-                                <li><a href="http://proxy.uchicago.edu/login?qurl={{'http://lib.uchicago.edu/h/worldcat'|urlencode:''}}"  data-ga-label="Search &gt; WorldCat">WorldCat</a></li>
-                                <li><a href="/search/catalogs/other" data-ga-label="Search &gt; Other Catalogs">Other Catalogs</a> <!-- Padding for lengthening column divider --> <p class="hidden-xs" style="padding-bottom:1em"></p></li>
+                                <li><a href="/h/3" data-ga-label="Library Catalog">Library Catalog</a></li>
+                                <li><a href="/h/borrowdirect" data-ga-label="BorrowDirect">BorrowDirect</a></li>
+                                <li><a href="http://proxy.uchicago.edu/login?qurl={{'http://lib.uchicago.edu/h/worldcat'|urlencode:''}}"  data-ga-label="WorldCat">WorldCat</a></li>
+                                <li><a href="/search/catalogs/other" data-ga-label="Other Catalogs">Other Catalogs</a> <!-- Padding for lengthening column divider --> <p class="hidden-xs" style="padding-bottom:1em"></p></li>
                             </ul>
                         </li>
                         <!-- // Search > Catalog column -->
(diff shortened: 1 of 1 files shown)


base/static/base/js/jquery.track-everything.js
```diff
@@ -253,19 +276,37 @@
...
-        const target = event.target.closest('a, button, input');
-        const eventName = getEventName(target);
-        if (!eventName) { return; }
+    function count_indecision(ep, eventName, target) {
+        // Function to count indecision clicks on tabs and dropdowns
+        // --- indecision count logic ---
+        if (eventName === 'tab' || eventName === 'dropdown') {
+            const component = target.closest('[data-ga-category]');
+            if (component) {
+                let count = parseInt(component.getAttribute('data-ga-indecision-count') || '0', 10) + 1;
```
bbusenius, line 285: If we put `if (isNaN(count)) count = 1;` here, we'll also handle for `NaN` values for cases when data might be corrupted.

```
  function count_indecision(ep, eventName, target) {
      // Function to count indecision clicks on tabs and dropdowns
      // --- indecision count logic ---
      if (eventName === 'tab' || eventName === 'dropdown') {
          const component = target.closest('[data-ga-category]');
          if (component) {
              let count = parseInt(component.getAttribute('data-ga-indecision-count') || '0', 10) + 1;
              if (isNaN(count)) count = 1; // <-- Just add this line
              component.setAttribute('data-ga-indecision-count', count);
          }
      }
  }
```

base/static/base/js/jquery.track-everything.js
```diff
@@ -276,17 +317,105 @@
...
+                        let optionsArr = currentOption ? currentOption.split('; ') : [];
+                        // Remove any previous type:... entry
+                        optionsArr = optionsArr.filter(opt => !opt.startsWith('type:'));
+                        if (selectedValue !== 'AllFields') {
+                            optionsArr.push('type:' + selectedValue);
+                        }
+                        currentOption = optionsArr.join('; ');
+                        searchButton.setAttribute('data-ga-event-option', currentOption);
+                    }
+                }
```
bbusenius, line 402: The logic for these could be refactored into handle functions, that could be called like so:
```
  function handleOptionChange(e) {
      const target = e.target;
      const searchWidget = target.closest(SELECTORS.SEARCH_WIDGET);
      if (!searchWidget) return;

      if (target.matches('input[type="checkbox"]')) {
          handleCheckboxChange(target, searchWidget);
      } else if (target.matches('input[type="radio"]')) {
          handleRadioChange(target, searchWidget);
      } else if (target.matches('select.selectpicker.btn-searchtype')) {
          handleSelectChange(target, searchWidget);
      }
  }
```

That might make this a little more maintainable. Not a big deal though.

bbusenius (approved):
I saw a couple areas for minor improvements but neither are worth blocking approval. Search tabs seem to be working fine and I don't see any problems. At a minimum, I'd at least add the `NaN` handling to the indecision count logic because it's only a 1 liner. I won't block on account of that though.

bbusenius (approved):
This looks good.

PR #856, opened 2025-09-10
854-proxy-link-converter-page

### LINKED ISSUE #854: proxy link converter page
The goal of this issue is to create a proxy link generator that is publicly accessible, containing an input field along with a button that changes the contents of that input field to be a proxified URL.

For example, if the input URL is:
https://doi.org/10.1632/BZCY1670

Then the output URL should be:
https://proxy-redirector-test.lib.uchicago.edu/login?url=https%3A%2F%2Fdoi.org%2F10.1632%2FBZCY1670

We will experiment with various UI possibilities, but one obvious choice is to create a route on the Wagtail site to something like `proxylinker` featuring this page.  Since there is a desire to embed this functionality within LibGuides, it may also be a good idea for the code to be easily copy-past-able into a LibGuides widget.

bufordrat:
Fixes #854.

The only change to the code is the addition of one javascript source file, `public/static/public/js/proxy-linker.js`, intended to be included in an html blob for embedding the proxy linker widget.

The page can be previewed here:
https://liblet.lib.uchicago.edu/borrow/access-privileges/offcampus/proxy-link-generator/


bbusenius (approved):
This looks good to me! Thanks for knocking this out!

vitorgomateus (changes requested):
```
pbutton.classList.remove("btn-secondary");
pbutton.classList.add("btn-primary");
```
to
```
pbutton.classList.remove("btn-primary");
pbutton.classList.add("btn-secondary");
cbutton.classList.remove("btn-secondary");
cbutton.classList.add("btn-primary");
```
So that the button that the user needs to click at any given moment, is the one that is red.

vitorgomateus (approved):
I meant to give those as suggestions and not stop the approval.

PR #860, opened 2025-09-29
857 improve custom events labeling

### LINKED ISSUE #857: Improve custom events labeling
The labeling of the custom events is not appropriate. Some elements in the main website still have inconsistent `category` definitions. Key pages in VuFind and Guides are not properly labeled at all. 

Prioritize adding HTML properties to elements instead of deducing values from context so that we can more easily create a script to overlay data fetched from GA on top.

For VuFind we need a hefty function that looks up elements and adds HTML `data-ga-*` properties for the tracking script. Might need for Guides as well. It would be nice for the Guides Widget on the main website as well since the content is produced manually in Wagtail instead of by code.

## New Schema 
- `name`
    - click (most things)
    - tab (for `role=”tab”`; and for the navbar dropdowns)
    - dropdown (`data-toggle="dropdown”`)
- `event_category`
    - Navigation
    - Main
    - Sidebar
    - Floating (Alert banners, feedback button)
- `event_subcategory`
    - Recognizable blocks,
    - blocks with similar concepts should share a word
        - widget
        - list
        - recent
- `event_label`
    - Exact text within the link OR role of the link. Where links are typically highly dynamic, the role of the link will be preferred
- `click_position`
- `event_indecision_count`
- `event_option`

### Special cases
The main search widget and the global navigation navbar have multiple clicks inside each, to help distinguish them, and to help the `indecision_count` function, these have a dedicated category each: "Main Search Widget" and "Global Navbar".

vitorgomateus:
Fixes #857 

**Changes in this request**
- introduce a function to inject HTML properties where it is difficult to work directly on the template.
- Introduce a better defined [schema](https://github.com/uchicago-library/library_website/wiki/Custom-Event-labeling) and updates templates according to it.
- *LibGuides and LibCal templates where update according to it.
- Add better context deduction for events on key VuFind pages: results and item views.
- Improve readability and documentation by grouping functions via object literals and moving documentation to a wiki.

**TODO**
- Testing and probable updates are still missing for the SFX interface.


base/static/base/js/jquery.track-everything.js
```diff
@@ -63,362 +29,572 @@
...
+                                                                    link.closest('.record-tab.details, .tab-pane.details-tab') ? 'Record Staff View' :
+                                                                        link.closest('.action-toolbar') ? 'Action Toolbar' :
+                                                                            'id:' + link.closest('[id]').getAttribute('id') || "VuFind Record Widget";
+                        params.event_label = link.closest('.savedLists') ? 'Record Saved in List' :
+                            link.closest('.bibToggle') ? 'More Details' : // How did VSCode knew to predict the value 'More Details' here?.
+                                link.classList.contains('title') ? 'Title' :
+                                    link.classList.contains('result-author') ? 'Author' :
+                                        link.classList.contains('external') ? 'Holding' :
+                                            link.classList.contains('save-record') ? 'Save Record' :
+                                                params.event_label || 'Unknown';
```
bbusenius, line 443: This is pretty fragile and it will break silently. I'm not sure if anything can be done about that, though. The nature of what you're trying to achieve here will likely be fragile by definition. I'm not sure how maintainable it will be, on that note. I guess there's no harm in trying though.

base/static/base/js/jquery.track-everything.js
```diff
@@ -63,362 +29,572 @@
...
-                    link.closest('.action-toolbar') ? 'Action Toolbar' :
-                    link.closest('.searchtools') ? 'Searchtools' :
-                        link.closest('.pagination') ? 'Pagination' :
-                            link.closest('[id]').getAttribute('id');
-                params.event_label = link.classList.contains('title') ? 'Title' :
-                    link.classList.contains('result-author') ? 'Author' :
-                        link.classList.contains('external') ? 'Holding' :
-                            link.classList.contains('save-record') ? 'Save Record' :
-                                params.event_label || 'Unknown';
+            if (rule.location && window.location.href.indexOf(rule.location) === -1) { return; }
```
bbusenius, line 147: Using `includes` in place of `indexOf` for these  `window.location.href` tests, might be a little cleaner and easier to read. `includes` is used for substring matching. Not a big deal if you want to go with `indexOf` though. I'm just throwing it out there:

```
if (rule.location && window.location.href.includes(rule.location) === -1) { return; }
```

vitorgomateus:
I'm pretty much adding everything @bbusenius suggested.

base/static/base/js/jquery.track-everything.js
bbusenius: Yeah, I agree. It's a difficult problem to solve. I appreciate keeping the template changes out of VuFind. We'll go with this for now and see how it works out.

base/static/base/js/jquery.track-everything.js
```diff
@@ -63,362 +29,533 @@
...
-            params.event_category = link.closest('[data-ga-category]').getAttribute('data-ga-category');
-        }
-        if (!params.event_subcategory && link.closest('[data-ga-subcategory]')) {
-            params.event_subcategory = link.closest('[data-ga-subcategory]').getAttribute('data-ga-subcategory');
-        }
-
-        if (link.closest('[data-ga-category][data-ga-indecision-count]')) {
-            params.event_indecision_count = link.closest('[data-ga-category][data-ga-indecision-count]').getAttribute('data-ga-indecision-count') || null;
+    // Rules for applyHtmlProperties
+    var htmlPropertyRules = [
```
bbusenius, line 50: Some parts of the code are using `ES5` variable declarations (as in this case) while other parts of the code are using `ES6` const/let patterns. We should update the script to always use `ES6` variable syntax. That would help with clarity and maintainability. Implementation would entail the following:

- Use `const` for values that won't be reassigned (configuration objects, DOM references after initial query)
- Use `let` for values that will be reassigned (loop counters, accumulator variables)

To update all of these, all you'll need to do is replace `var` with `const` (when the variable isn't reassigned) or `let` if the variable changes.

Example with `const`:

```
  const htmlPropertyRules = [
      {
          location: LOCATIONS.LIB,
          selector: '#widget-featured-library-expert',
          childSelector: 'a',
          attribute: 'data-ga-label',
          value: "Expert's Subjects",
      },
      ...
  ];
```

Example with `let`:

```
  let headingText = '';
  if (rule.useFirstHeading) {
      if (typeof rule.useFirstHeading === 'object' && Array.isArray(rule.useFirstHeading.levels)) {
          headingText = getFirstHeadingTextWithin($items, rule.useFirstHeading.levels);
      } else {
          headingText = getFirstHeadingTextWithin($items);
      }
  }
```

base/static/base/js/jquery.track-everything.js
```diff
@@ -63,362 +29,533 @@
...
-                params.event_label = link.closest('.s-srch-result-guide') ? 'Guide Name' :
-                    link.closest('.s-srch-result-author') ? 'Guide Author' :
-                        link.closest('.s-srch-result-subjects') ? 'Guide Subject' :
-                            link.closest('.s-srch-result-url') ? 'Guide Link' :
-                                link.closest('.s-srch-result-title') ? 'Guide Page Title' :
-                                    link.closest('.s-lg-label-more') ? 'More Button' : params.event_label || 'Unknown';
+        rules.forEach(function (rule) {
+            if (!rule || !rule.selector || !rule.attribute) { return; }
+
+            if (rule.location && window.location.href.includes(rule.location) === -1) { return; }
```
bbusenius, line 107: Nice!

base/static/base/js/jquery.track-everything.js
```diff
@@ -63,362 +29,533 @@
...
+                    let optionsArr = currentOption ? currentOption.split('; ') : [];
+                    // Remove any previous type:... entry
+                    optionsArr = optionsArr.filter(opt => !opt.startsWith('type:'));
+                    if (selectedValue !== 'AllFields') {
+                        optionsArr.push('type:' + selectedValue);
+                    }
+                    currentOption = optionsArr.join('; ');
+                    searchButton.setAttribute('data-ga-event-option', currentOption);
+                }
+            }
```
bbusenius, line 238: Good example of `ES6` syntax that I mentioned in the other comment. Since we're doing this in parts of the code, we should just do it everywhere and not use `var`.

base/static/base/js/jquery.track-everything.js
```diff
@@ -63,362 +29,533 @@
...
+        },
+        count_indecision(ep, eventName, target) {
+            // Function to count indecision clicks on tabs and dropdowns
+            // --- indecision count logic ---
+            if (eventName === 'tab') {
```
bbusenius, line 243: The old ` count_indecision` function counted dropdowns too, this one doesn't. I just want to be sure this is intentional. Disregard if that's what you want. I'm just checking.

bbusenius (changes requested):
This is looking great! I just requested a couple of small changes.

PR #871, opened 2025-10-15
Optimize test suite performance and fix parallel test issues

### LINKED ISSUE #868: Intermittent test failure: test_normal_page_shows_in_sitemap_xml
## Description

Two intermittent test failures occur related to Wagtail's site URL resolution returning `None`:

1. `test_normal_page_shows_in_sitemap_xml` in `public/tests.py`
2. `test_news_article_page_routing_to_category_listing` in `lib_news/tests.py`

### Issue 1: public/tests.py

```
AssertionError: None != 'http://starfleet-academy.com/the-great-link-test/'
```

The sitemap URL location is `None` instead of the expected URL.

### Issue 2: lib_news/tests.py

```
File "/app/lib_news/models.py", line 294, in base_url
    return self.get_url_parts()[-1]
TypeError: 'NoneType' object is not subscriptable
```

The `get_url_parts()` method returns `None` when trying to resolve the base URL.

### Issue 1: Missing request parameter in production code

The `PublicBasePage.get_sitemap_urls()` method in `base/models.py:2234` is not passing the `request` parameter to its parent class method:

```python
return super(PublicBasePage, self).get_sitemap_urls()  # Missing request!
```

Without the request object, Wagtail cannot determine the site context and thus cannot generate proper URLs.

**Proposed Fix**: Pass the `request` parameter:
```python
return super(PublicBasePage, self).get_sitemap_urls(request)
```

### Issue 2: Stale cache in tests

Both test failures are also caused by stale cache data between tests. As explained in `base/tests.py:246-248`:

> "Explicitly clear the cache of site root paths. Normally this would be kept in sync by the Site.save logic, but this is bypassed when the database is rolled back between tests using transactions."

When Django's TestCase rolls back the database between tests, Site objects are removed but multiple cache layers still contain stale references:
- `wagtail_site_root_paths` cache (Django cache)
- Django URL resolver caches
- Wagtail page cache (wagtailcache)

This causes `get_url_parts()` to return `None` intermittently.

**Proposed Fix**: Clear all cache layers in both `setUp()` and `tearDown()`:

```python
from django.core.cache import cache
from django.urls import clear_url_caches
from wagtailcache.cache import clear_cache

def setUp(self):
    # Clear cache from previous tests
    clear_url_caches()
    cache.clear()
    clear_cache()
    # ... rest of setup

def tearDown(self):
    # ... cleanup
    clear_url_caches()
    cache.clear()
    clear_cache()
```

This pattern is used in `LinkQueueSpreadsheetBlockTestCase` (base/tests.py:756-760) and ensures a clean cache state across test boundaries.

bbusenius:
# Optimize Test Suite Performance

Fixes #868

This PR improves test suite performance through multiple optimizations while also fixing several bugs that were causing issues in parallel test execution.

### Before:
- **Parallel (16 cores):** ~10 minutes
- **Sequential:** ~14 minutes

### After:
- **Parallel (16 cores):** **3m 25s** (-66%, -6m 35s)
- **Sequential:** **7m 1s** (-50%, -7 min)

### 1. Converted Test Classes to Use setUpTestData()

**What changed:** Converted 10 test classes (44 tests total) from using setUp() to setUpTestData().

**Why this matters:** 
- [setUp()](cci:1://file:///home/brad/Documents/Business/UChicago/Code/library_website/units/tests.py:6:4-20:29) runs before **every single test method**, recreating all test data each time
- [setUpTestData()](cci:1://file:///home/brad/Documents/Business/UChicago/Code/library_website/ask_a_librarian/tests.py:0:4-1:45) runs **once per test class**, creating data in a database transaction that's shared across all test methods in that class
- Django automatically rolls back the transaction after the class finishes, so tests remain isolated
- For test classes that create complex page hierarchies (sites, pages, staff, units, etc.), this eliminates massive redundant work

**Example impact:** A test class with 11 test methods that previously created the same 10 pages 11 times now creates them once.

**Classes optimized:**
- News page and category tests (5 tests)
- Collection area page tests (11 tests)
- Exhibit filtering tests (5 tests)
- Staff page supervisor tests (1 test)
- Staff listing tests (2 tests)
- Standard page exclude fields tests (4 tests)
- Standard page sitemap tests (2 tests)
- Link queue spreadsheet block tests (6 tests)
- Quick number utils tests (6 tests)
- Ask utils tests (2 tests)

### 2. Removed Low-Value Smoke Test

**What changed:** Deleted `test_all_live_public_pages_for_200_or_redirect_with_anonymous_user` which took 304 seconds.

**Why this matters:**
- This test made individual HTTP requests to every single live page in the test database (hundreds of pages)
- It only verified status codes (200, 301, or 302) without checking actual functionality
- Individual page type tests already provide better coverage by testing specific features
- The test was essentially a slow, brittle smoke test that didn't catch real bugs

**Impact:** Removed 304 seconds of test time with no loss in meaningful coverage.

### 3. Optimized Collections Filter Tests

**What changed:** Reduced exhaustive iteration tests from 202 total requests to 11 representative samples with edge cases.

**Why this matters:**
- Previous tests iterated through ALL subjects (118), ALL locations (73), and ALL formats (11)
- Each iteration made a full HTTP request, queried the database, and rendered templates
- All iterations tested the same code path - the filtering logic doesn't change based on which subject/location you pass
- New approach: Test a few representative cases + invalid parameter handling

**Results:**
- Subject filtering: 87s → 3.3s (-96%)
- Location filtering: 82s → 2.6s (-97%)
- Format filtering: Optimized to 3 samples
- Added new tests for invalid parameter handling (better coverage!)
- Net result: Added 3 new tests while dramatically reducing execution time

### 4. Fixed Parallel Test Bugs

**Issues fixed:**
- `get_sitemap_urls()` was missing required request parameter in method signature
- `relative_url()` can return None in certain parallel test scenarios, causing crashes
- Added null checks to handle these edge cases gracefully
- Removed `localhost` site from test fixture and added automatic cleanup in setup scripts to prevent site conflicts that were causing intermittent parallel test failures

**Why this matters:** Tests now run reliably in parallel without random failures.

### 5. Infrastructure Improvements

**Test environment cleanup:**
- Removed `localhost` site from test fixture to prevent conflicts
- Added automatic `localhost` site deletion to Vagrant and Docker setup scripts
- Updated test helper functions to ensure clean, consistent test environments
- These changes were critical for eliminating intermittent failures in parallel test execution

**CI optimization:**
- Removed `--keepdb` flag from GitHub Actions workflow
- This flag is only useful for local development where the database persists between runs
- In CI, every run starts with a fresh environment, so `--keepdb` provides no benefit

**Output cleanup:**
- Suppressed unnecessary `stdout` in test commands for cleaner, more readable output

## Testing

All 114 tests pass in both parallel and sequential modes. The test suite is now significantly faster while maintaining the same level of coverage and reliability.

### DIFF
--- a/ask_a_librarian/tests.py
+++ b/ask_a_librarian/tests.py
@@ -1,16 +1,21 @@
+from base.tests import add_generic_request_meta_fields
+from base.utils import get_hours_and_location
 from django.http import HttpRequest
 from django.test import TestCase
+from library_website.settings import (
+    CRERAR_HOMEPAGE,
+    DANGELO_HOMEPAGE,
+    DISSERTATION_HOMEPAGE,
+    ECKHART_HOMEPAGE,
+    PUBLIC_HOMEPAGE,
+    SCRC_HOMEPAGE,
+    SSA_HOMEPAGE,
+)
+from public.models import PublicRawHTMLPage, StandardPage
 from wagtail.models import Site
 
 from ask_a_librarian.models import AskPage
 from ask_a_librarian.utils import get_chat_statuses, get_unit_chat_link
-from base.tests import add_generic_request_meta_fields
-from base.utils import get_hours_and_location
-from library_website.settings import (
-    CRERAR_HOMEPAGE, DANGELO_HOMEPAGE, DISSERTATION_HOMEPAGE, ECKHART_HOMEPAGE,
-    PUBLIC_HOMEPAGE, SCRC_HOMEPAGE, SSA_HOMEPAGE
-)
-from public.models import PublicRawHTMLPage, StandardPage
 
 
 class TestAskUtils(TestCase):
@@ -18,11 +23,9 @@ class TestAskUtils(TestCase):
     # See base.tests to learn how to create this file
     fixtures = ['test.json']
 
-    def setUp(self):
-        """
-        Common stuff used in multiple test cases.
-        """
-        self.ask_pages = AskPage.objects.all()
+    @classmethod
+    def setUpTestData(cls):
+        cls.ask_pages = AskPage.objects.all()
 
     def test_get_chat_statuses(self):
         """
@@ -31,8 +34,9 @@ class TestAskUtils(TestCase):
         """
         chat_statuses = get_chat_statuses()
         self.assertEqual(
-            len(self.ask_pages), len(chat_statuses),
-            'There are AskPages not present in the get_chat_statuses function'
+            len(self.ask_pages),
+            len(chat_statuses),
+            'There are AskPages not present in the get_chat_statuses function',
         )
 
     def test_get_unit_chat_link(self):
@@ -42,9 +46,7 @@ class TestAskUtils(TestCase):
         and SCRC are treated differently since they're circumstances
         are slightly different.
         """
-        ask_widgets = set(
-            ['law', 'crerar', 'ssa', 'uofc-ask', 'dissertation-office']
-        )
+        ask_widgets = set(['law', 'crerar', 'ssa', 'uofc-ask', 'dissertation-office'])
 
         # Dictionary of tuples where the keys map to the ask_widget_name field
         # of AskPages. The firs item of the tuple is a mixed dictionary of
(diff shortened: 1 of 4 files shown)


bufordrat (approved):
This is awesome!  Tests pass on my machine.  I don't know how you were able to figure out so many different little problems all at once.

bbusenius:
Thanks @bufordrat. Claude helped a lot. I think `setUpTestData` was the big one.

PR #923, opened 2025-11-10
Unit field validation error due to legacy limit_choices_to constraint

### LINKED ISSUE #918: Unit field validation error due to legacy limit_choices_to constraint
## Problem

When selecting a Unit for a StandardPage (or any page extending `PublicBasePage`), users encounter a validation error:

```
Select a valid choice. That choice is not one of the available choices.
```

This occurs when the selected UnitPage has `display_in_dropdown=False`.

## Root Cause

**Location:** `base/models.py:1578-1585`

The `unit` field in `PublicBasePage` has a legacy constraint:

```python
unit = models.ForeignKey(
    'units.UnitPage',
    null=True,
    blank=False,
    limit_choices_to={'display_in_dropdown': True},  # <-- Legacy constraint
    on_delete=models.SET_NULL,
    related_name='%(app_label)s_%(class)s_related',
)
```

## Why This Happens

1. **Historical Context:** The `display_in_dropdown` field was designed to filter units when they appeared in dropdown menus in older versions of Wagtail.

2. **Current Behavior:** Wagtail now uses PageChooser widgets for ForeignKey fields to pages, which displays *all* UnitPages in the chooser modal, completely ignoring the `limit_choices_to` filter.

3. **The Mismatch:** Users can select any UnitPage via the PageChooser interface, but Django's form validation still enforces the `limit_choices_to` constraint on submission. If the selected unit has `display_in_dropdown=False`, validation fails.

Image

## Proposed Solution

Remove the `limit_choices_to={'display_in_dropdown': True}` constraint from the `unit` field in `base/models.py` line 1582, since:

1. It no longer serves its original purpose (filtering dropdown choices)
2. It creates a confusing UX where users can select options that will be rejected on save
3. PageChooser provides adequate selection UI without needing this filter

This same constraint exists throughout the codebase science other page types inherit from `base`. The source of truth is in `base/models.py`.

## Impact

Low risk change - removes an outdated constraint that's causing validation errors. No migration needed since `limit_choices_to` is only enforced at the Django form level, not at the database level (double check this).

bbusenius:
Fixes #918

## Summary

Removes the outdated `limit_choices_to={'display_in_dropdown': True}` constraint from the `unit` ForeignKey field in `PublicBasePage`. This constraint causes validation errors when users select units with `display_in_dropdown=False` via the PageChooser interface.

- Removed `limit_choices_to` constraint from `unit` field in base/models.py:1638
- The constraint no longer serves its original purpose since Wagtail's PageChooser ignores it
- Eliminates confusing UX where selectable options are rejected on form submission

## Migration Status

⚠️ **Database migrations will be added after PR #912 merges** to avoid migration number conflicts in the `lib_collections` app. Once #912 is merged, we'll need to:
1. Sync this working branch with the latest master
2. Run `docker compose exec web ./manage.py makemigrations`
3. Push the generated migration files

The `limit_choices_to` parameter only affects Django form validation, not database schema, so these will be no-op migrations at the database level.

## Testing

After migrations are generated and applied:
1. Edit a StandardPage (or any page extending PublicBasePage) in the Wagtail admin
2. Use the PageChooser to select a Unit that has `display_in_dropdown=False`
3. Save the page
4. Verify the page saves successfully without validation errors


bufordrat (approved):
Fantastic!  This false-positive validation error has been a hassle during testing over the past couple years; great to get rid of it.

PR #930, opened 2025-11-18
Non-admin users with editor permissions cannot save edits to StaffPages

### LINKED ISSUE #929: Non-admin users with editor permissions cannot save edits to StaffPages
## Problem

Non-admin users with editor permissions are unable to save edits to StaffPages. When attempting to save, they receive a validation error message even when only editing fields they have permission to access (Content tab).

## Error

The page displays: "The page could not be saved due to validation errors"

The underlying issue is that formset validation is failing with "ManagementForm data is missing" errors for the InlinePanels in the HR tab (staff_page_email, staff_page_phone_faculty_exchange, staff_page_units).

## Context

This issue was introduced in PR #865 which added the `permission='staff.change_staff_hr_info'` parameter to the HR tab's ObjectList. While this correctly hides the HR tab from users without HR permissions, it causes Django formset validation to fail because the hidden InlinePanels don't render their management forms in the HTML.

## Expected Behavior

Users with editor permissions (but without HR permissions) should be able to edit and save StaffPages. The HR data should remain unchanged when these users save pages.

bbusenius:
Fixes #929

## Summary

Non-admin users with editor permissions were unable to save edits to StaffPages due to formset validation errors. PR #865 added permission restrictions to hide the HR tab, but this caused Django to expect formset management forms that weren't being rendered.

- Created custom StaffPageForm that dynamically removes HR fields when user lacks HR permissions
- Removes formsets not present in POST data before validation runs
- HR data remains unchanged when saved by non-HR users

## Testing

1. Create or use a test user with editor permissions but without `staff.change_staff_hr_info` permission
2. Log in as that user and navigate to edit a StaffPage
3. Verify the HR tab is not visible
4. Make changes to fields in the Content tab (e.g., update bio or profile picture)
5. Save the page
6. Verify the page saves successfully without validation errors
7. Verify HR data (cnetid, email addresses, phone numbers, units) remains unchanged

### DIFF
--- a/staff/models.py
+++ b/staff/models.py
@@ -13,6 +13,7 @@ from modelcluster.fields import ParentalKey
 from rest_framework import serializers
 from subjects.models import Subject
 from units.models import BUILDINGS
+from wagtail.admin.forms.pages import WagtailAdminPageForm
 from wagtail.admin.panels import (
     FieldPanel, HelpPanel, InlinePanel, MultiFieldPanel, ObjectList, PageChooserPanel,
     TabbedInterface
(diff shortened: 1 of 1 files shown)


vitorgomateus (approved):
Works on my local!

PR #952, opened 2026-01-07
StaffPage crashes with AttributeError after removal of is_active field from GroupPage

bbusenius:
Fixes #951

Summary

After PR #942 removed the is_active field from GroupPage, StaffPage.get_context() crashes with AttributeError when trying to filter group memberships. This fix removes the is_active check and implements hierarchical group membership display.

- Remove is_active check from StaffPage.get_context() that was causing AttributeError
- Organize group memberships hierarchically by GroupIndexPage ancestors using get_ancestors().type(GroupIndexPage)
- Update template to display groups in hierarchical structure with index page headers
- Move GroupIndexPage import to module level for cleaner code

Testing:
1. Navigate to any staff page on the intranet (e.g., http://loopdev:8000/staff/[staff-member]/)
2. Verify the page loads without errors
3. Check that "Committees & Groups" section displays groups organized under their parent GroupIndexPage names
4. Verify that all group memberships are shown (both current and past)

PR #955, opened 2026-01-09
StaffPage displays "Inactive Groups" before "Committees & Groups" in hierarchical group listing

### LINKED ISSUE #954: StaffPage displays "Inactive Groups" before "Committees & Groups" in hierarchical group listing
## Issue

After PR #952 implemented hierarchical group memberships on StaffPage, the GroupIndexPages are displayed in an unpredictable order. In production, "Inactive Groups" appears before "Committees & Groups", which is confusing since active groups should be listed first.

Additionally, the individual groups within each section are not alphabetized, making it difficult to scan long lists (especially "Inactive Groups" which can have 20+ items).

## Root Cause

In `staff/models.py:617`, the `grouped_memberships` dictionary is converted to a list without any sorting:

```python
hierarchical_group_memberships = list(grouped_memberships.values())
```

The order depends on which GroupIndexPage is encountered first when iterating through group memberships, making it unpredictable. Similarly, the groups within each section are in the order they were encountered in the loop.

## Proposed Solution

1. Sort the GroupIndexPages alphabetically by title, which will naturally place "Committees & Groups" before "Inactive Groups"
2. Sort the individual groups within each section alphabetically by title for easier scanning

```python
# Sort groups within each section
for group_data in grouped_memberships.values():
    group_data["groups"].sort(key=lambda x: x["title"])

# Sort sections by index page title
hierarchical_group_memberships = sorted(
    grouped_memberships.values(),
    key=lambda x: x["index_page"]["title"]
)
```

## Notes

- This solution is somewhat fragile as it relies on alphabetical ordering of titles
- Works for the current intranet use case with just these two GroupIndexPages
- A more robust solution would require adding explicit ordering fields to GroupIndexPage, but that's overkill for this simple case

bbusenius:
Fixes #954

Summary

After PR #952, GroupIndexPages were displayed in unpredictable order on StaffPages. This adds alphabetical sorting by index page title, ensuring "Committees & Groups" appears before "Inactive Groups". Additionally, the individual groups within each section are not alphabetized, making it difficult to scan long lists (especially "Inactive Groups" which can have 20+ items).
- Sort hierarchical_group_memberships by index page title using sorted() with key function
- Sort groups within group categories.

Testing:
1. Go to `nest` and ensure that "Committees & Groups" comes before "Inactive Groups" and that groups are alphabetically sorted in each category: https://looplet.lib.uchicago.edu/staff/elizabeth-edwards/

### DIFF
--- a/staff/models.py
+++ b/staff/models.py
@@ -613,10 +613,13 @@ class StaffPage(BasePageWithoutStaffPageForeignKeys):
                     }
                 )
 
+        # Sort groups within each section alphabetically
+        for group_data in grouped_memberships.values():
+            group_data["groups"].sort(key=lambda x: x["title"])
+
         # Convert dict to list for template and sort by index page title
         hierarchical_group_memberships = sorted(
-            grouped_memberships.values(),
-            key=lambda x: x["index_page"]["title"]
+            grouped_memberships.values(), key=lambda x: x["index_page"]["title"]
         )
 
         context = super(StaffPage, self).get_context(request)


bufordrat (approved):
Awesome.

PR #1014, opened 2026-04-14
IntranetIndexPage child page listing ignores Wagtail menu sort order

### LINKED ISSUE #1013: IntranetIndexPage child page listing ignores Wagtail menu sort order
## Problem

On `IntranetIndexPage` pages (e.g. https://loop.lib.uchicago.edu/documentation/folio/), the auto-generated list of child pages in the body always displays alphabetically by title. Editors using Wagtail's "Sort menu order" action to reorder children see their changes saved, but the reordering has no effect on the rendered page.

Reported in [Bugzilla #32524](https://trouble.lib.uchicago.edu/bugzilla/show_bug.cgi?id=32524).

## Cause

In `base/models.py`, `IntranetIndexPage.get_context()` builds the child page tree from `self.get_children().live()` (which is in Wagtail menu/path order), then explicitly re-sorts it alphabetically before rendering:

```python
def alphabetize_pages(currentlevel):
    for node in currentlevel:
        node["children"] = alphabetize_pages(node["children"])
    return sorted(currentlevel, key=lambda c: c["title"])

pages = alphabetize_pages(pages)
```

This affects both the flat listing and the `display_hierarchical_listing` mode.

## Proposed fix

Remove the `alphabetize_pages` call so the listing respects Wagtail's menu order (which is what editors expect from the "Sort menu order" action). If alphabetical sorting is desired for some pages, consider making it an opt-in `BooleanField` on the page model similar to `display_hierarchical_listing`.

## Testing

1. Go to an `IntranetIndexPage` with several child pages in the Wagtail admin.
2. Use "Sort menu order" to put the children in a non-alphabetical order and save.
3. Visit the page on the front end and confirm the listing reflects the chosen order.

bbusenius:
Fixes #1013

Summary

The auto-generated child page list on `IntranetIndexPage` was being re-sorted alphabetically by title, which silently discarded any reordering editors made with Wagtail's "Sort menu order" action.

- Remove the `alphabetize_pages` helper and its call in `IntranetIndexPage.get_context()` so the listing falls through to Wagtail's default `get_children()` order (menu/path order)
- Affects both the flat listing and the `display_hierarchical_listing` mode

Testing:
1. Pick an `IntranetIndexPage` with several child pages (e.g. the FOLIO Loop page at `/documentation/folio/`).
2. In the Wagtail admin, use "Sort menu order" to put the children in a non-alphabetical order and save.
3. Visit the page on the front end and confirm the listing reflects the chosen order rather than alphabetical order.

### DIFF
--- a/base/models.py
+++ b/base/models.py
@@ -2423,13 +2423,6 @@ class IntranetIndexPage(BasePage):
                 )
             )
 
-        def alphabetize_pages(currentlevel):
-            for node in currentlevel:
-                node["children"] = alphabetize_pages(node["children"])
-            return sorted(currentlevel, key=lambda c: c["title"])
-
-        pages = alphabetize_pages(pages)
-
         def get_html(currentlevel):
             if not currentlevel:
                 return ""


vitorgomateus (approved):
Looks good!

bufordrat (approved):
Nice!  Works well in dev mode on my machine.

PR #1015, opened 2026-04-16
add custom event tracking properties

vitorgomateus:


bbusenius (approved):
This looks great! This was a lot more like what I was expecting. All of the CI jobs are passing for me locally too. I had Claude do a review and it flagged a few issues with the labels, however, I don't actually know if the feedback is warranted because it looks to me like some of it is intentional. Nevertheless, I'll drop the feedback here for you to look at, just in case some of it is relevant. I don't see any problems though.

```
  TabNav.js (line 95-96): data-ga-category={tab.label} sets category to the tab's own label, and data-ga-label={tab.label} duplicates it. Categories should be stable      
  groupings (e.g., "Tabs" or "Main"), not per-item labels — setting both to the same value provides no extra dimension to GA, and it overrides the data-ga-category="Main" 
  on #mylib-dashboard for every tab click.                                                                                                                                 
                                                                                                                                                                           
  MyAccountSidebar.js: Two different links carry data-ga-label="Saved Items" (lines ~64 and ~257) and two carry data-ga-label="Fines" (lines ~87 and ~102). In GA they'll  
  collapse into a single label so you can't tell which rendered variant was clicked — the fines warning variant especially should be distinguishable from the plain one    
  (e.g., "Fines (warning)").                                                                                                                                               
                             
  AutoRenewalNotice.js: data-ga-category="Floating" introduces a new category that isn't in the CATEGORIES constants in base/static/base/js/jquery.track-everything.js     
  (Navigation/Footer/Sidebar/Main/etc.). Intentional, but worth verifying with Vitor that it should live outside the existing taxonomy.
                                                                                                                                                                           
  CategoryCard.js: data-ga-subcategory={${title} Card} is fine, but the two manageLabel links (top and bottom of the card) will emit identical category+subcategory+label  
  tuples — click position won't let you tell top vs bottom manage-link clicks apart.
```

bbusenius (approved):
This looks good to me.

PR #1023, opened 2026-05-13
Automate Out of Sync Staff Report Updates

### LINKED ISSUE #1022: Automate Out of Sync Staff Report Updates
## Background

Every day, the `list_staff_wagtail --report-out-of-sync-staff` management command runs in cron and emails a report of differences between Wagtail's staff records and the University's campus directory API. Today, every diff line in that report is handled manually: a web team member reads the email and clicks through the Wagtail admin to update names, position titles, emails, phone numbers, faculty exchanges, departments — plus the chore of creating users + pages for new hires and deactivating users for departures.

Most of these changes are mechanical and could be automated. The report already does all the hard work: it queries the directory API, computes set differences against Wagtail, and emits a structured list of every disagreement keyed by CNetID and field.

## Proposal

Add a new `sync_staff_with_directory` management command (in `staff/management/commands/`) that consumes the same `WagtailStaffReport._staff_out_of_sync()` output the report does and applies the obvious updates automatically. Schedule it in cron to run **before** the existing report each day; anything it doesn't (or can't) handle still surfaces in the report email as it does today, so the script can never cause silent data loss — at worst it makes a bad edit that shows up in tomorrow's email.

### Behavior per CNetID

The report emits two lists. By parsing each line of the form `"{cnetid} -{value}- ({field})"` (plus the bare `{cnetid}` lines that signal a new hire or departure), every CNetID falls into one of three buckets:

**1. Field-level update on an existing staff member**
- Scalar fields (`officialName`, `displayName`, `positionTitle`) — overwrite the Wagtail value with the API value.
- A `displayName` change cascades to `StaffPage.title`, the derived `first_name` / `middle_name` / `last_name` on the StaffPage, the `first_name` / `last_name` on the matching Django `User`, and `StaffPublicPage.title`.
- Multi-value fields (`email`, `phoneFacultyExchange`, `department`) — remove anything the report flags as Wagtail-only, add anything it flags as API-only. The directory is authoritative.
- Departments resolve by matching the report's string against `UnitPage.get_campus_directory_full_name()`. If any one department in the diff can't be resolved, skip all department updates for that person and let the next report flag it for manual handling.

**2. New hire** (bare CNetID on the API side)
- Create a Django `User` with the `Library` + `Editors` groups, reusing the existing `create_library_user` management command so that logic stays in one place.
- Create a `StaffPage` under the existing staff parent (page id 38), populated from the report's API-side values.
- Create a `StaffPublicPage` under the existing public parent (page id 4273). Default `editor`, `page_maintainer`, and `content_specialist` to the newly created `StaffPage` itself (self-managed); default `unit` to the top-level `Library` UnitPage. If no `Library` UnitPage can be found, skip the public page and let the next report surface it.

**3. Departure** (bare CNetID on the Wagtail side)
- Set `user.is_active = False` and save. The existing `post_save` hook at `staff/wagtail_hooks.py:112` already unpublishes both the StaffPage and the StaffPublicPage when a user goes inactive — we just need to fire the signal programmatically.

### Design choices to confirm

- **No dry-run mode.** The daily out-of-sync report email *is* the dry-run; it tells us exactly what will be applied. Adding a separate dry-run mode is redundant.
- **Plain `.save()` rather than `save_revision().publish()`** for updates to existing pages. `save_revision()` snapshots the page's in-memory modelcluster state, and `publish()` then reverts direct `.create()` / `.delete()` calls on child rows (emails and phone-FX). This matches the precedent at `base/management/commands/update_staff_first_and_last_names.py`. New pages still need `save_revision().publish()` to be published in the first place.
- **CNetID changes** look to the report like one departure + one new hire. We accept that — actual CNetID changes happen roughly once per decade, and the report will catch the misclassification anyway.
- **No API sanity gate.** The existing `_staff_out_of_sync()` already calls the directory; we inherit its behavior. We've never seen a zero-cnetid day, and the script can't do anything worse than the daily report could already mislead us into doing manually.
- **`--cnetid <id>`** flag for limiting a run to a single person — useful for smoke-testing against a real diff before turning the script loose in cron.

### Per-CNetID transactions

Each CNetID is processed in its own `transaction.atomic` block. A failure on one person logs an `[cnetid] ERROR: ...` line and moves on; the rest of the run still completes.

## Out of scope

A few cleanup opportunities surfaced while planning this that could be follow-up issues rather than blockers:

- Refactor `WagtailStaffReport._staff_out_of_sync()` to return structured records instead of formatted strings, so the new command (and any future consumer) doesn't have to reverse-parse `_format()`'s output via regex.
- Extract `create_library_user.Command`'s body into a plain function in `base/utils.py` so the new command can call it directly rather than via `call_command()` with stringy `"False"` / `"True"` booleans.
- There's a stray `staff_page.page_maintainer = staff_page; staff_page.save()` at `base/management/commands/create_or_update_library_staff_page.py:91-92` that's a no-op — `StaffPage` doesn't have a `page_maintainer` field (it extends `BasePageWithoutStaffPageForeignKeys`). Worth deleting in a separate cleanup.

## Acceptance criteria

- [ ] `staff/management/commands/sync_staff_with_directory.py` exists and runs cleanly with `--help`.
- [ ] Optional `--cnetid` flag limits a run to one person.
- [ ] Scalar field updates apply correctly, including the `displayName` cascade across `StaffPage`, `StaffPublicPage`, and `User`.
- [ ] Email and phone-FX rows are added / removed to match the API set; multi-row staff records are handled correctly.
- [ ] Departments resolve against `UnitPage.get_campus_directory_full_name()`; unresolvable departments are skipped + logged without breaking other field updates.
- [ ] New hires create a `User` (in `Library` + `Editors`), a populated `StaffPage`, and a `StaffPublicPage` with sane editor / maintainer / content_specialist / unit defaults.
- [ ] Departures call `user.save()` with `is_active=False`; the existing hook unpublishes both pages.
- [ ] Per-CNetID atomic transactions; logs go to stdout for cron to email.
- [ ] Unit tests cover the parsing helpers and each branch (scalar update, displayName cascade, email replacement, the `dakotaw`-style phone-FX upgrade, department add, department resolution failure, departure, retire idempotency, new-hire flow, the `--cnetid` filter).

bbusenius:
Fixes #1022

### Summary

Adds a `sync_staff_with_directory` management command that consumes `WagtailStaffReport._staff_out_of_sync()` output and applies the diffs automatically. Intended to run in cron just before the existing report, so anything it misses still surfaces in tomorrow's email.

- Field updates (officialName/displayName/positionTitle/email/phoneFacultyExchange/department); displayName change cascades to StaffPage, StaffPublicPage, and User
- New hires create User (Library + Editors), StaffPage, and StaffPublicPage with self-referenced editor/maintainer/content_specialist and Library as unit
- Departures set `user.is_active = False`; existing `post_save` hook handles unpublishing both pages
- Per-CNetID atomic transactions; `--cnetid` flag for single-record smoke tests; 23 unit tests covering helpers and every branch

### Testing
This is on `nest`, you can play with it there. You can edit the HR tab for people on their `StaffPages` to make them out of sync.
- Run `./manage.py list_staff_wagtail --report-out-of-sync-staff` to see what would be in the out of sync staff report.
- Run `./manage.py sync_staff_with_directory` to actually do the syncing. After you're done, if you run `list_staff_wagtail --report-out-of-sync-staff` again, it won't report any changes.

### DIFF
--- a/staff/management/commands/sync_staff_with_directory.py
+++ b/staff/management/commands/sync_staff_with_directory.py
@@ -176,7 +176,9 @@ class Command(BaseCommand):
             bundles = {k: v for k, v in bundles.items() if k == target}
 
         if not bundles:
-            self.stdout.write("No diffs to apply.")
+            # Nothing to sync: stay silent so the daily cron run produces no
+            # output (and thus no email) on quiet days. Cron only mails us
+            # when this command actually changes records or hits an error.
             return
 
         self._unit_by_campus_name = {
--- a/staff/management/commands/sync_staff_with_directory.py
+++ b/staff/management/commands/sync_staff_with_directory.py
@@ -1,0 +1,487 @@
+# -*- coding: utf-8 -*-
+"""
+Apply the diffs from `list_staff_wagtail --report-out-of-sync-staff` to Wagtail.
+
+Intended to run from cron just before the existing out-of-sync report. Anything
+this command can't or doesn't auto-handle will still surface in the next
+report email and can be fixed manually.
+"""
+
+from __future__ import unicode_literals
+
+import re
+
+from django.contrib.auth.models import User
+from django.core.management import call_command
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from wagtail.models import Page
+
+from base.models import make_slug
+from public.models import StaffPublicPage
+from staff.models import StaffPage
+from staff.utils import WagtailStaffReport
+from units.models import UnitPage
+
+# Page tree parents under which new pages are created. These ids are stable
+# fixtures of the site and match the parents used when creating staff manually
+# in the Wagtail admin.
+STAFF_PAGE_PARENT_ID = 38
+STAFF_PUBLIC_PAGE_PARENT_ID = 4273
+
+# Format produced by `WagtailStaffReport._staff_out_of_sync()._format`:
+#   "{cnetid} -{value}- ({field})"
+LINE_RE = re.compile(r"^(?P<cnetid>\S+) -(?P<value>.*)- \((?P<field>[a-zA-Z]+)\)$")
+PHONE_RE = re.compile(r"(\d{3}-\d{3}-\d{4})$")
+
+SCALAR_FIELDS = ("officialName", "displayName", "positionTitle")
+MULTI_FIELDS = ("email", "phoneFacultyExchange", "department")
+
+
+def parse_phone_fx(value):
+    """
+    The report flattens (faculty_exchange, phone_number) into one space-joined
+    string. Split it back: phone (if present) is anchored at end-of-string,
+    everything else is the faculty exchange.
+    """
+    m = PHONE_RE.search(value)
+    if m:
+        return value[: m.start()].strip(), m.group(1)
+    return value.strip(), ""
+
+
+def split_name(display_name):
+    """
+    Split a display name into (first, middle, last) by whitespace.
+
+    Args:
+        display_name: a single-line name from the directory's displayName,
+            e.g. "Pete Sample" or "Joan Q. Public".
+
+    Returns:
+        A 3-tuple of strings. Rules:
+            - 0 tokens -> ("", "", "")
+            - 1 token  -> (token, "", "")
+            - 2 tokens -> (first, "", last)
+            - 3+ tokens -> (first, " ".join(middles), last) — covers e.g.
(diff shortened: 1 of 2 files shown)


bufordrat (approved):
Glorious.  Tried both the management command and the test suite out on `nest`, and everything worked perfectly.

PR #1026, opened 2026-05-22
861 create an ags spreadsheet upload page

### LINKED ISSUE #861: create an AGS spreadsheet upload page
Currently, Find It! works with information about which articles are available via Article Galaxy hardcoded into it.  The goal of this issue is to create an upload page in Loop that Document Delivery staff can use to upload the spreadsheet with this information that the Article Galaxy web app provides.  Wagtail will then process that information and serve it up live as a JavaScript source file, which can be fetched onto `sfx.lib` in `cron` and src-included in Find It!

The workflow will be:

- Document Delivery staff upload the new `.xlsx` file through an upload button on a Loop page specifically for this purpose
- after uploading, they are able to preview the `.xlsx` file directly on the page to quickly vet the new data
- Wagtail will save a backup copy of the original `.xlsx` file to the database
- Wagtail will pull all the information it needs out of the spreadsheet and arrange it into the form of a JSON object
- it will serve this JSON object up in a view on the public site, which, to the outside world, will look like a static JavaScript file

bufordrat:
Fixes #861.

This PR adds an upload page for Document Delivery staff to use, located at `/ags`.  The page exists as a straight Django view within the `intranethome` app.  

Workflow for Document Delivery staff:

- staff member navigates to `/ags`
- they upload the latest spreadsheet export from the Article Galaxy application
- the page validates the spreadsheet
- if it fails validation, the staff member is given an error message explaining the problem and asked to try again
- if it passes validation, a Wagtail Document called `ags_spreadsheet.xlsx` is created in the database
- this document is the basis for the `/ags/js` view, which displays the information in the spreadsheet as a lookup table representing journal coverages
- this allows Find It to look up a journal's ISSN, then determine whether its year is within the coverage range for the journal answering to that ISSN

Putting this feature into production is step 1.  The second step, which we will pursue separately, is to create a `cron` job on `sfx.lib` that does a curl to `/ags/js` to pull down the latest JSON every few minutes, which Find It will source as a `.js` file.

# Testing

To run unit tests:

```console
$ ./manage.py test intranethome.tests.test_ags_upload
```

To manually test:

- navigate to `/ags`
- upload the test file located at `intranethome/test_data/export.xlsx`
- a spreadsheet preview should appear in the page at `/ags`
- navigate to `/ags/js` to see the dictionary that will be sent over to `sfx.lib`
- to delete the spreadsheet, hit the delete button at `/ags`
- `/ags/js` will now display error JSON
- save a local copy of `export.xlsx` with a fake row added and one row deleted
- upload that new file to the page to see the diff preview display

# Style

This PR uses monadic error handling.  Utility code to help with monadic error handling has been added to a module called `base/result.py`.


base/utils.py
```diff
@@ -468,6 +473,47 @@ def generator(tup):
...
+    Helper function for determining whether the user is authorized
+    to view a page.
+
+    Args: http request, wagtail page
+
+    Returns: bool
+    """
+    groups = get_required_groups(page)
+    user = request.user
+    return has_permission(user, groups)
```
bbusenius, line 487: Nice use of existing functions! Good utility to have!

base/utils.py
```diff
@@ -468,6 +473,47 @@ def generator(tup):
...
+
+def get_loop_homepage():
+    """
+    Helper function for retrieving Loop page.
+
+    Args: none
+
+    Returns: wagtail page
+    """
+    return Site.objects.get(site_name="Loop").root_page
```
bbusenius, line 498: There are other ways of doing this, including using not and the `default_site` property (or something like that), or putting the site ID in config. I think we do it differently in different cases in the codebase, which isn't great. Anyhow, don't change this, I'm just pointing it out. I think it's great to have this as a base utility. I just did something like this in my last PR (not as a base utility), and I kind of wish I had done this instead. I like it.

base/utils.py
```diff
@@ -468,6 +473,47 @@ def generator(tup):
...
+
+    Args: http request
+
+    Returns: http response or none
+    """
+    loop_homepage = get_loop_homepage()
+    if not has_page_permissions(request, loop_homepage):
+        return redirect_users_without_permissions(loop_homepage, request, None, None)
+    else:
+        return response
```
bbusenius: Just a question: I wonder if this is similar enough to `redirect_users_without_permissions` in `base/wagtail_hooks.py` that we could update the latter to rely on whatever you do here? Just a thought.

intranethome/templates/intranethome/ags_upload_page.html
```diff
@@ -0,0 +1,120 @@
...
+<div>
+    <p />
+    <span style="color: #499172;">
+	{{ greens }} subscriptions added
+    </span>
+    ;
+    <span style="color: #c73c4f;">
+	{{ reds }} subscriptions removed
+    </span>
+    <p />
```
bbusenius: We should probably address this spacing with CSS instead of the empty `<p />`s.

intranethome/ags.py
```diff
@@ -0,0 +1,421 @@
...
+    Convert XLSX Wagtail Document into pandas dataframe.
+
+    Args:
+        doc: Wagtail document object.
+
+    Returns:
+        Spreadsheet as a pandas dataframe.
+    """
+    with doc.file.open() as f:
+        return handle_to_df_exn(f)
```
bbusenius, line 196: Beautiful!

intranethome/templates/intranethome/ags_upload_page.html
```diff
@@ -0,0 +1,120 @@
...
+</div>
+{% endif %}
+
+<title> Article Galaxy Configuration Management </title>
+
+<h1> Article Galaxy Configuration Management </h1>
+
+<div>
+    <form action="/ags/" method="post" enctype="multipart/form-data" id="ags_form" onsubmit="return confirm('Are you sure you want to update the AGS spreadsheet?')">
+	<input type="hidden" name="csrfmiddlewaretoken" value="6Cy5s5Ohq8RAQhZJkr1Yh2Zrb7J7z8kOktLobd2S8PZ87In8zy7LMLg9Ba1KLnH8">
```
bbusenius: I think we should get rid of this line (23). `{% csrf_token %}` handles it.

intranethome/views.py
```diff
@@ -1,19 +1,35 @@
...
+    diff_rows,
+)
+from base.utils import (
+    permissions_redirect,
+    get_loop_homepage,
+    has_page_permissions,
 )
+from django.template.response import TemplateResponse
+from django.db.utils import ProgrammingError, OperationalError
+from django.shortcuts import render, redirect
```
bbusenius: `OperationalError`, `ProgrammingError` and `render` are imported twice. I don't think we need these later ones (and `redirect` can move up). This is probably some of what are causing the linting errors.

intranethome/ags.py
```diff
@@ -0,0 +1,421 @@
...
+
+    Args:
+        request: An HTTP request.
+
+    Returns:
+        binary XLSX data
+    """
+    if request.method == "POST" and "uploadFile" in request.FILES:
+        upload_file = request.FILES["uploadFile"]
+        with upload_file.open(mode="rw") as f:
```
bbusenius: Claude was really serious about `rw` not being valid here and that it should be `r+` or `rb`. I made him provide me with a citation for that and pushed back and got him to repent because I didn't interpret the it the same way. Only flagging it in case you might have a different interpretation: https://docs.python.org/3/library/functions.html#open

intranethome/views.py
```diff
@@ -259,4 +269,114 @@ def mail_aliases_view(request, *args, **kwargs):
...
 
-    return render(request, "intranethome/mail_aliases.html", context)
+    return permissions_redirect(
+        request,
+        render(request, "intranethome/mail_aliases.html", context)
+    )
+
+
+def ags_upload_page(request):
+
```
bbusenius, line 291: It would be a good idea to add something like:

```
if not has_page_permissions(request, get_loop_homepage()):
    return redirect_users_without_permissions(get_loop_homepage(), request, None, None)
```

here so that `create_document(AGS_SPREADSHEET_NAME)` can't ever run for an authenticated user. That being said though, this is just a stylistic concern, because `csrf` token protection and our loop permissions redirection are actually preventing an unauthorized submission here, so it's not technically a security bug. But you did something similar in `delete_spreadsheet` and I think it would be a good idea to do the same here. Not a blocker though.

bbusenius (changes requested):
This is a nice and thoughtful pull request. I very much appreciate the functional programming style and the monadic error handling. This is something I didn't know anything about and I enjoyed learning about it. It's clear and useful. It would be nice to expand this approach to other things. I also think your handling of diffing spreadsheets is a thoughtful and nice touch. It works well and looks good when rendered. It's a nice addition!

Most of my feedback is just comments about things I like or very minor tweaks. The main issues are these:

1. The `test_data` wasn't included, which is why tests are failing.
2. Linting jobs are failing. There are some repeat imports I noticed. This is surely part of it. I'm guessing there are other little things but should be easily resolved.

Other more minor things I noticed:
1. I have a question about whether `permissions_redirect` can be used to update `redirect_users_without_permissions` in `base/wagtail_hooks.py`. At a glance, it looks like they could and should be coupled. I'm not 100% on this though, so. I'll defer to your judgement.
2. We should remove an unnecessary manual dev `csrf` token.
3. We should probably address the empty `<p />` spacing with CSS instead.
4. It would be nice to add a permissions check at the top of `ags_upload_page` in `views.py` similar to what you did in `delete_spreadsheet`. This is not a true security hole but would be a nice extra protection.

Honestly, this is strong work and a nice addition. If CI is passing, I think it can be merged.

base/utils.py
bufordrat: Interesting suggestion; I haven't looked at this part of the code before.  What does the `@hooks.register("before_serve_page")` decorator on top of `redirect_users_without_permissions` do, exactly?  I'm guessing it just calls this function, passing the current page/http request/arguments in, upon page load, but under what circumstances does that happen?  Is it for every page, or only for certain pages?  My recollection is that without an explicit call to this function, the site wouldn't block me from accessing a Loop page when not authenticated.

intranethome/templates/intranethome/ags_upload_page.html
bufordrat: Good call; I think I had it in the back of my head to do that and then forgot.  How do you usually handle these things; do you create a `div` and then add a certain class to it for vertical spacing?

intranethome/templates/intranethome/ags_upload_page.html
bufordrat: Eeek!  Not sure how that got in there; thanks.  Could be a copy/paste error from during development.

intranethome/templates/intranethome/ags_upload_page.html
bbusenius: There are lots of ways to do it, all valid. In this case I might just throw classes on the spans and then write rules for them. You can also use the CSS descendant selector, if it's relevant (scoping down through the parent classes, e.g. `div.something_unique span`  will style all spans within a div with a class of `something_unique`. But creating `div`s here, is also totally valid.

intranethome/views.py
bufordrat: Oops; thanks.

bufordrat: Must have forgotten to run the autoformatters etc. on this file.

base/utils.py
bbusenius: The hook fires before any Wagtail page is served, not just Loop pages, but since `get_required_groups` returns an empty list for pages with no restrictions, `has_permission` returns `True` and nothing happens. Loop pages have a `PERMISSIONS_MAPPING` such that causes unauthorized users (anyone not in the Library group) to get redirected. You're correct that you need to make an explicit call to this function because you're in a view, not a page (in a page the hook handles it automatically). 

What I was wondering was if there was a DRY way we could integrate this with `redirect_users_without_permissions`, however, my question was flawed because that's literally what you're doing here. I missed that somehow. Consider this resolved.

intranethome/ags.py
bufordrat: Yikes, total brain fart.  Should have been `rb`.  Great catch.

intranethome/views.py
bufordrat: Hmm; great point.  I hadn't considered that the code leading up to the return statement might get run before the permissions check happened.  This underscores the need for a better api for permissions_redirect.  Will experiment with that a bit.

intranethome/ags.py
bbusenius: I really wasn't sure, to be honest. Claude was saying that it worked because Django uses `UploadedFile.open()` which ignores the `w` but that it would use `TemporaryUploadedFile` on a large file, which would throw a value error. I didn't fully believe it. Shows what I know. LOL.

intranethome/views.py
bufordrat: Ah, I remember now why I gave it the API of returning an HTTP response.  The trouble is that if you just have it be a helper function, it's not clear how to get it to do an early return, which is the normal way we do redirects.

bufordrat: ok, trying something new out...

bbusenius (approved):
Looks good! I think this will be a nice addition.

PR #1033, opened 2026-06-23
MyLib Dashboard: paged items in "In transit" don't appear in any tab

### LINKED ISSUE #1032: MyLib Dashboard: paged items in "In transit" don't appear in any tab
## Summary

A user reported that of two paged UChicago items in FOLIO, one (`Open - Awaiting pickup`) correctly shows under **Available for Pickup**, but the other — currently **in transit** — does not appear anywhere on the dashboard, including the **In Process** tab where they expected it.

Reported after the deploy of #958 (MyLib Dashboard).

## Root cause

The dashboard splits a patron's requests across two tabs using two narrow, exact-match status queries:

- `get_user_holds()` matches **only** `status=="Open - Awaiting pickup"` → *Available for Pickup* tab.
  - `mylib_dashboard/services/folio.py:469-473`
- `get_paging_requests()` matches **only** `status=="Open - Not yet filled"` (and `requestType==Page or Hold`) → *In Process* tab.
  - `mylib_dashboard/services/folio.py:664-670`

A FOLIO Page/Hold request moves through these states:

```
Open - Not yet filled  →  Open - In transit  →  Open - Awaiting pickup  →  Closed - Filled
```

`Open - In transit` (item retrieved, being transported to a pickup point that differs from its effective location — common for cross-library UChicago paging) matches **neither** query, so the request is invisible until it reaches the hold shelf. `Open - Awaiting delivery` falls in the same gap.

The deeper problem is the approach: *In Process* is defined by an **enumerated whitelist** of statuses. Any open status not on that list disappears — that's how this bug happened, and any status FOLIO adds later will hit it too.

## Fix: define In Process by complement, not by whitelist

There is exactly one open request status that is **not** in process: `Open - Awaiting pickup` (the item is on the hold shelf, ready). Everything else that's open is still being fulfilled. So partition the patron's open requests against that single status:

- **Available for Pickup** = open requests where `status == "Open - Awaiting pickup"` (unchanged)
- **In Process** = open requests where `status != "Open - Awaiting pickup"`

This covers `Not yet filled`, `In transit`, `Awaiting delivery`, and any future open status automatically, and the two tabs become a clean, mutually-exclusive split of the patron's open requests.

This mirrors what the VuFind FOLIO driver already does (UChicago runs it in production). `getMyHolds()` fetches **all** open requests with a single wildcard query rather than per-status queries, then sorts by status — `module/UChicago/src/UChicago/ILS/Driver/Folio.php:675`:

```php
$query = ['query' => '(' . $userQuery . ' and status == Open*)'];
```

## Proposed implementation

1. Change `get_paging_requests()` to fetch open requests (`status==Open*`) and exclude `Open - Awaiting pickup` in Python, rather than matching a single hard-coded status. Avoid relying on a CQL not-equals in the query string — fetch-all-then-filter is the path VuFind took.
   - `mylib_dashboard/services/folio.py:664-670`
2. Drop the `requestType==Page or Hold` restriction. Under an "any open status" rule that filter reintroduces the same kind of gap (e.g. an open **Recall** the patron placed would be dropped), and the *Available for Pickup* query has no requestType filter, so removing it keeps the two tabs symmetric. **Judgment call — confirm before implementing.**
3. Surface the request status in the UI so a patron can distinguish "being retrieved" from "in transit." The backend already returns `status` per request, but `PagingRequestItem` never renders it.
   - `mylib_dashboard/static/mylib_dashboard/js/components/PagingRequestItem.js`
4. Update test coverage to exercise the full set of open statuses (`In transit`, `Awaiting delivery`, etc.), not just `Open - Not yet filled`.
   - `mylib_dashboard/tests/test_services_folio.py`

## How to reproduce

1. As a patron, place two page requests for items whose effective location differs from the chosen pickup location.
2. Let one reach the hold shelf → it shows under *Available for Pickup* (correct).
3. Observe the other while it is mid-transit (`Open - In transit`) → it appears in neither *Available for Pickup* nor *In Process*.

bbusenius:
Fixes #1032

Summary

The MyLib Dashboard In Process tab queried only requests with status "Open - Not yet filled", so a paged item that had progressed to "Open - In transit" (or "Open - Awaiting delivery") showed up in neither In Process nor Available for Pickup — it disappeared from the dashboard until it reached the hold shelf. This selects every open status except "Open - Awaiting pickup" so in-flight requests stay visible.

- `get_paging_requests()` now fetches all open requests (`status==Open*`) and excludes only "Open - Awaiting pickup" (which belongs to Available for Pickup); the `requestType==Page or Hold` filter is unchanged
- Show the FOLIO request status as a badge on each In Process item (`Open - ` prefix stripped for display)
- Added service tests covering in-transit/awaiting-delivery inclusion and awaiting-pickup exclusion

Testing:
1. Place a page request for an item whose home location differs from the pickup location, and let it progress to "In transit" in FOLIO (or set a test request to status "Open - In transit").
2. Load the MyLib Dashboard and open the In Process tab — the item appears with an "In transit" status badge.
3. Confirm an "Awaiting pickup" request still shows only under Available for Pickup, not In Process.
4. `docker compose exec web ./manage.py test mylib_dashboard.tests.test_services_folio.TestGetPagingRequests`


vitorgomateus (approved):
I can't really test this fully right now, but looks solid.

PR #1037, opened 2026-07-02
CGIMail Form Editor broken: OpenAI retired the chatgpt-4o-latest model

### LINKED ISSUE #1036: CGIMail Form Editor broken: OpenAI retired the chatgpt-4o-latest model
## Problem

The CGIMail Form Editor page (Rebecca) fails on every generate request with:

```
Error: OpenAI API error: 404 Client Error: Not Found for url: https://api.openai.com/v1/chat/completions
```

## Cause

OpenAI retired the `chatgpt-4o-latest` model from the API on February 17, 2026 ([deprecations page](https://developers.openai.com/api/docs/deprecations)). The page's `ai_model` field defaults to `chatgpt-4o-latest` and the existing page has that value stored, so every request now returns `model_not_found` (which OpenAI reports as a 404).

Related problems found while investigating:

1. **Most of the other model choices are also being retired.** Of the six options on `CGIMailFormEditorPage.ai_model`, only `gpt-5.1` survives past this year: `gpt-4o`, `gpt-4-turbo`, and `gpt-3.5-turbo` shut down October 23, 2026, and `o3` shuts down December 11, 2026. The page will break again in a few months if we only swap out the dead default.
2. **Current models reject our request payload.** `generate_form_json` sends `temperature: 0.2` for non-o1/o3 models, but GPT-5 series models only accept the default temperature, so simply pointing the page at a current model returns a 400. The o1/o3 special-casing (no system message) is also obsolete — current models all accept system messages.
3. **The error handler hides the real error.** The view returns `str(e)` from `requests`, which only shows the HTTP status line. OpenAI's response body says exactly what's wrong (`model_not_found`, `insufficient_quota`, etc.) but it never reaches the user, which made this failure look like a broken endpoint rather than a retired model.

## Proposed fix

- Replace the `ai_model` choices with current models (e.g. `gpt-5.5` as default, plus `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.3-chat-latest`) and migrate the field.
- Simplify the OpenAI payload in `generate_form_json`: always send a system message, drop the `temperature` parameter and the o1/o3 branch.
- Surface the error message from the OpenAI response body in the view's error handler.
- Update the existing page in the Wagtail admin to use one of the new models (the stored `chatgpt-4o-latest` value is not changed by the field migration).

## Testing

On the page, describe a form and click generate. It should return generated form JSON instead of the 404 error.

bbusenius:
Fixes #1036

Summary

The CGIMail Form Editor failed on every generate request because OpenAI retired the chatgpt-4o-latest model from the API on February 17, 2026, and most of the other configured model choices are scheduled for shutdown later this year. This replaces the model lineup with current models and fixes the request payload and error reporting so the page works with them and future API failures explain themselves.
- Replaced the `ai_model` choices on `CGIMailFormEditorPage` with current models: `gpt-5.5` (new default), `gpt-5.4`, `gpt-5.4-mini`, and `gpt-5.3-chat-latest`; migration included
- Simplified the OpenAI payload in `generate_form_json`: always send a system message and drop `temperature: 0.2` and the o1/o3 branch, since GPT-5 series models only accept the default temperature and all current models accept system messages
- Error handler now surfaces the message from the OpenAI response body (e.g. `model_not_found`, `insufficient_quota`) instead of just the HTTP status line
- Updated tests for the new payload shape and added a test for the error message pass-through

Testing:
Run the tests: `docker compose exec web ./manage.py test cgimail_editor` (all 22 pass; no live API calls). Then migrate, open the CGIMail Form Editor page in the Wagtail admin, set the AI model to one of the new choices (the stored `chatgpt-4o-latest` value is not changed by the migration), publish, and generate a form on the page — it should return form JSON instead of the 404 error.


vitorgomateus (approved):
works great!

bufordrat (approved):
Totally working in dev on my machine, as near as I can make out.

PR #1056, opened 2026-07-28
Add a "Page Listing" StreamField block for dynamic child-page indexing (sitewide, public + Loop)

### LINKED ISSUE #1051: Add a "Page Listing" StreamField block for dynamic child-page indexing (sitewide, public + Loop)
## Summary

Add a new `DefaultBodyFields` StreamField block, **"Page Listing"**, that renders a dynamic, nested index of a page's descendant pages. This gives "index page powers" to (almost) any page on both the public site and Loop, and is modeled on the "Dynamic Page Listing" feature from the Research Data site that the team liked — reworked as a block instead of a page-level feature.

In practice this makes the intranet's dedicated index page types redundant, but we are **not** removing them in this issue (see Out of Scope).

## Background

The Research Data site implements "Dynamic Page Listing" as a **page-level** feature: two fields (`show_nested_children`, `nested_children_depth`) plus a recursive template tag that renders a nested `<ul>` of live, in-menu descendants rooted at the current page. It's simple and well-liked, but being page-level it isn't portable, can't be positioned within content, and can't appear more than once.

This site already has **four separate, copy-pasted variants** of "list my descendants as a nested list":

1. `IntranetIndexPage.get_context` — `base/models.py:2389+` (unlimited depth or flat; builds HTML strings in Python)
2. `PublicBasePage` `enable_index` / `display_hierarchical_listing` — `base/models.py:176-231`, `2162-2193` (public analog; also builds HTML in Python)
3. "Browse this Section" left sidebar — `base/models.py:1467-1484` (2 levels)
4. `IconAutoListBlock` — `icon_list_boxes/models.py:26-61` (2 levels; **already a StreamField block** and the closest existing precedent)

They're inconsistent about `.in_menu()`/`.specific()`, depth caps, and URL handling. This block will supersede *some* of them — notably the intranet index page types — but not necessarily all: the left-sidebar "Browse this Section" nav, for example, serves a different purpose and likely stays separate. The list above is here mainly to give an overview of related functionality already in the codebase.

## Proposal

A new `StructBlock` (working name `PageListingBlock`, editor label **"Page Listing"**) defined in `base/models.py` alongside the other blocks, added as a child of `DefaultBodyFields` (`base/models.py:1278`) in the `group="Layout and Data"` group.

Because `IntranetDefaultBodyFields` subclasses `DefaultBodyFields` (`base/models.py:1402`), the block flows to Loop pages automatically — a single point of change covers both sites.

### Block fields

- **`heading`** — optional `CharBlock`. Editor-controlled section heading (Research Data hard-coded "Further Reading"; we let editors set or omit it).
- **`root_page`** — optional `PageChooserBlock`. When blank, defaults to the current page (the "index of this page" behavior, i.e. the `IntranetIndexPage` replacement). When set, lists that section instead — any section anywhere, not just a descendant of the current page.
- **`depth`** — `ChoiceBlock` (levels 1–3, default `1`, where 1 = immediate children only). Hard-capped at **3**.

### Behavior

- Lists **live** pages that are **in menus** (`show_in_menus`) — respecting `show_in_menus` gives editors a lever to hide individual pages from listings.
- Nesting rendered as a nested `<ul>`, depth-limited to the selected value (hard-clamped to the cap).
- Default tree/menu ordering (the manual order set in the page explorer).

### Implementation notes

- **Query:** fetch descendants in a **single query** — `root.get_descendants().live().in_menu().filter(depth__lte=root.depth + levels)` — and build the tree in Python. Avoid the naive recursive `get_children()` per node (N+1) used by the existing variants; some Loop sections are large. The hard depth cap is what keeps the single query bounded.
- **Rendering:** use a **recursive template partial** (`base/templates/base/blocks/page_listing.html` + an included node partial that includes itself). Do **not** build HTML strings in Python + `|richtext` (the anti-pattern in variants 1 and 2).
- **Current page:** resolve via `get_context(self, value, parent_context=None)` → `parent_context['page']` when `root_page` is blank.
- **Coverage:** flows to every page using `DefaultBodyFields`/`IntranetDefaultBodyFields`. It will **not** appear on `TOCPage` (uses `IconListCluster`) or the one page using `RawHTMLBodyField` — acceptable.
- **Migrations:** adding a block to `DefaultBodyFields` generates an auto-migration for each model using it (~15 page models across ~12 apps). These are **state-only** StreamField definition changes (JSON column) — no DDL, no data migration — but it does touch many apps.
- Add tests: rendering, depth clamping at the cap, `show_in_menus`/`live` filtering, blank-vs-set `root_page`.

## Decisions

The open questions below have been resolved and are settled for v1:

1. **Root page — ship the optional `root_page` chooser.** Allowing any section anywhere gives the most functionality and avoids having to move pages or content around when, for example, a working group concludes its work but the parent group/department still wants to display its outputs.
2. **`show_in_menus` — respect it.** Only live, in-menu pages are listed.
3. **Depth cap — 3.** Well structured documentation can nest deeper, but the cap has to land somewhere and 3 is sufficient.
4. **Display extras — no.** v1 is a plain nested text index: no per-item summaries/excerpts, thumbnails, or card layout. Open to reconsidering if editors find they want more options; precedent exists in Research Data's separate `SectionBlock` (card grid with thumbnails/excerpt/count) if we revisit.

## Out of scope (possible follow-ups)

- Removing or deprecating `IntranetIndexPage` and other intranet index page types — they stay for now.
- Auditing the other listing variants above to decide which, if any, should later be folded into this block (the sidebar in particular is a different use case and probably stays as-is).
- Card/thumbnail/summary display modes; custom ordering; page-type filtering; pagination.

## Acceptance criteria

- [ ] "Page Listing" block available in the body of public and Loop pages that use `DefaultBodyFields`.
- [ ] Renders a nested, depth-limited list of live, in-menu descendants (default: current page as root).
- [ ] Optional `root_page` chooser lists any section anywhere when set.
- [ ] Depth selectable in the editor (1–3) and hard-capped server-side at 3.
- [ ] Single-query descendant fetch; recursive template partial for rendering.
- [ ] Tests covering depth cap, filtering, and root selection.
- [ ] Migrations generated and committed.

bbusenius:
Fixes #1051

## Summary

Adds a "Page Listing" StreamField block that renders a dynamic, nested index of a page's descendants, giving "index page powers" to almost any page on the public site and Loop. Because `IntranetDefaultBodyFields` subclasses `DefaultBodyFields`, adding the block in one place covers both sites.

- New `PageListingBlock` in `base/models.py`, added to `DefaultBodyFields` under "Layout and Data": optional `heading`, optional `root_page` chooser, and a `depth` choice of 1-3.
- Lists **live**, **in-menu** pages in page explorer order. Leaving `root_page` blank lists the children of the current page; setting it lists any section anywhere.
- Depth is hard-capped server-side at 3 in `build_page_listing()`, so the cap holds regardless of what a stored value says. A page excluded from the listing takes its branch with it.
- Descendants come from a single query (`get_descendants().live().in_menu().filter(depth__lte=...)`) assembled into a tree in Python on the treebeard path — no per-node `get_children()`. Rendering uses a recursive template partial instead of building HTML strings in Python.
- 12 state-only migrations across the apps using `DefaultBodyFields`. StreamField definition changes to the JSON column: no DDL, no data migration.

### Note for reviewers: the `{% include_block %}` change

Every body StreamField on this site rendered as `{{ self.body }}` or `{% for block in self.body %}{{ block }}{% endfor %}`. Both call `StreamValue.__html__()`, which takes no arguments, so `Block.render()` receives `context=None` and `parent_context` is never populated. Per [Wagtail's docs](https://docs.wagtail.org/en/stable/topics/streamfield.html), `parent_context` "is available when the block is rendered through an `{% include_block %}` tag" — the short form "does not pass variables from the calling template such as `request` or `page`" and is recommended only "for simple values that do not render HTML of their own."

A block cannot discover the page it sits on any other way: the block instance is a single shared definition reused for every page and request, and the stored value is just `{heading, root_page, depth}`. So the blank-`root_page` default — the `IntranetIndexPage` replacement, and what an editor gets by default — required switching those render sites to `{% include_block %}`.

Output is unchanged. The per-child loops became `{% include_block block %}` rather than collapsing to a whole-stream `{% include_block self.body %}`, which would have added Wagtail's `<div class="block-*">` wrappers to markup whose CSS never expected them. `wagtailcore_tags` was already loaded in every affected template.

### One editor-facing gotcha

Wagtail's `show_in_menus` defaults to `False`, and the block respects it (decision #2 in the issue). `IntranetIndexPage` never filtered on it, so a Loop section whose pages don't have "Show in menus" checked will list nothing where the old index page listed everything. Sections with working left-sidebar nav are unaffected, since that nav already uses `.in_menu()`.

Out of scope per the issue: removing `IntranetIndexPage` and the other intranet index page types, auditing the other listing variants, and card/thumbnail/summary display modes.

## Testing

Full suite passes: 373 tests, including 11 new ones in `base.tests.PageListingBlockTestCase`.

    docker compose exec -T web python manage.py test base.tests.PageListingBlockTestCase

They cover the depth cap (including a bogus stored depth), `live`/`show_in_menus` filtering and branch exclusion, nesting structure, blank vs. set `root_page` rendered end-to-end through `page.serve()`, the heading, an empty section rendering nothing, and an `assertNumQueries(1)` guard on the descendant fetch.

To try it by hand:

1. Pick a page with a few levels of children beneath it and confirm the children have "Show in menus" checked.
2. Add a **Page Listing** block to its body under "Layout and Data", leave **Root page** blank, set **Depth** to 2 or 3, publish.
3. The page should show a nested `ul.index-list` of its descendants. Unchecking "Show in menus" or unpublishing a child should drop that page and everything under it.
4. Set **Root page** to an unrelated section and confirm the listing follows the chosen section instead.
5. Repeat on a Loop page to confirm the block is available and styled there.
6. Spot-check a few page types whose templates changed — a standard page, a news story, a staff bio, an exhibit, a collection — to confirm existing body content renders as before.


bufordrat (approved):
Fantastic.  Works perfectly on my machine in dev.  I also really like the way the validation on adding a loop page tree worked out.

PR #1069, opened 2026-08-18
Pin pillow-heif to <1.5 for FreeBSD libheif 1.22 compatibility

### LINKED ISSUE #1068: Pin pillow-heif to <1.5 for FreeBSD libheif 1.22 compatibility
## Problem

When deploying on FreeBSD with libheif 1.22.x, `pip install` fails building `pillow-heif` because versions >= 1.5.0 require libheif >= 1.23.1, which is not available on FreeBSD.

```
pillow_heif/_pillow_heif.c:6:6: error: "pillow_heif requires libheif >= 1.23.1"
```

## Solution

Pin `pillow-heif` to `>=1.4.0,<1.5` in `requirements.txt`. Version 1.4.0 is the last release that only requires libheif >= 1.19.0, which is compatible with FreeBSD's libheif 1.22.x.

bbusenius:
Fixes #1068

Summary

pillow-heif 1.5.0+ requires libheif >= 1.23.1, which is unavailable on FreeBSD 14, causing wheel builds to fail during deployment. Pins pillow-heif to >=1.4.0,<1.5 — the last series compatible with libheif 1.22.x.
- Added `pillow-heif>=1.4.0,<1.5` to `requirements.txt`

Testing:
Run `pip install -r requirements.txt` on a FreeBSD system with libheif 1.22.x and confirm it installs successfully.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,6 +25,7 @@ git+https://github.com/bbusenius/django-turnstile-site-protect.git#egg=django_tu
 git+https://github.com/bbusenius/Good-Bots.git
 html2text>=2025.4.15,<2026
 pandas>=2,<3
+pillow-heif>=1.4.0,<1.5
 pocli==0.1.10
 psycopg2
 requests[socks]


bufordrat (approved):
Very nice.  Nothing iffy that I can think of or see in dev.

